PR #11, opened 2023-12-30
Implement retriever and doc manager as agents

dbrewster:
Implemented doc retriever as an agent

Implemented doc manager and fixed up all of the parsers.


sdk/eidos_sdk/agent_os.py
```diff
@@ -10,12 +10,14 @@ class AgentOS:
     file_memory: "FileMemory" = ...  # noqa: F821
     symbolic_memory: "SymbolicMemory" = ...  # noqa: F821
     similarity_memory: "VectorMemory" = ...  # noqa: F821
+    embedder: "Embedding" = ...  # noqa: F821
```
LukeLalor, line 13: Is the embedder a property of similarity memory or a machine characteristic?

sdk/eidos_sdk/memory/agent_memory.py
```diff
@@ -281,21 +278,45 @@ async def query(
             )
         return returnDocuments
 
+    async def raw_query(
```
LukeLalor, line 281: What is this needed for? it looks like the only user grabs the embeddings the same way the normal query does. If it is the query item vs document wrapper, perhaps we should soothe over those differences instead.

sdk/pyproject.toml
```diff
@@ -35,6 +35,7 @@ unstructured = "^0.11.2"
 python-multipart = "^0.0.6"
 pillow = "^10.1.0"
 aiohttp = "^3.9.1"
+watchdog = "^3.0.0"
```
LukeLalor, line 38: watchdog only looks like it is used in examples, should it be in the sdk?

LukeLalor (commented):
Some questions. It would also be good if we had some end some agent tests which hit a few parsers / similarity memory.

thanks for the rest unit tests on the pdf parser. Those should also hit similarity memory?

PR #14, opened 2024-01-03
flatten references, start changing examples

LukeLalor:
This pr changes the reference structure so it does not need to be nested

Before
```
ref_key:
    implementation: f.q.n
    spec:
        arg1: val1
        arg12 val2
```

After
```
ref_key:
    implementation: f.q.n
    arg1: val1
    arg12 val2
```

It also bounds the resource types. Now all resources are Machine, Agent, or Reference. Because of that agents now need to reference their implementation (ie, GenericAgent, TreeOfThoughtAgent, or fqn) since it can not be retrieved from kind for free.

I also changed how we define builtins so that they are normal yaml definitions which can be copied/overiden without much thought. As such we can also refer to a named Reference when defining a default for a Reference within a spec.


sdk/eidos_sdk/builtins/generic_agent.yaml
LukeLalor, line 1: I threw this in for your concept of easily overriden defaults. This is a clear way to define builtin system resource which customers can define themselves to override. In the AnnotatedReference we can also define the default as a string pointing to a builtin resource. In another pr we will migrate existing defaults to builtins to be explicit.

sdk/eidos_sdk/system/resources.py
LukeLalor, line 1: I split this file up to help deal with python imports

dbrewster (approved):

PR #16, opened 2024-01-05
Ll/easily override builtins

LukeLalor:


dbrewster (approved):
Looks good 

I think we should clean up some of the names but we can do that after you push.

PR #18, opened 2024-01-06
build docker images

LukeLalor:
Fixed some bugs in our local symbolic memory to allow code searcher to work with it.

add git workflows to publish base image (and break up existing publish sdk workflow)

image layers should be pretty good. We are installing all dependencies before dropping our wheel on, so we preserve dependency cache unless they are updated.

decided do install from wheel vs pypi to leave it decoupled from pypi publishing. Should be identical.


sdk/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG BASE_IMAGE=python:3.11-slim
```
LukeLalor, line 1: todo, I should change the default here to be the secure one dave was talking about

LukeLalor: Punting on this one. Iron bank doesn't have a container registry

dbrewster (approved):

PR #27, opened 2024-01-10
fun group conversations

dbrewster:


sdk/eidos_sdk/cpu/agent_cpu.py
```diff
@@ -69,25 +71,31 @@ def __init__(self, call_context: CallContext, cpu: AgentCPU):
...
+    ) -> T:
+        if get_origin(output_format) == list:
+            class Wrapper(BaseModel):
+                items: output_format
+
+            rtn = await self.schedule_request(prompts, Wrapper)
+            return rtn.items
+
+        model = TypeAdapter(output_format)
+        schema = model.json_schema()
```
LukeLalor, line 93: I think we need to do a type check here in case we have an actual schema dict, then we can't convert it to a TypeAdapter before hand

LukeLalor: we can do it via "get_python_type" from you schema_to_model module.

In this scenario we probably still want to validate, but return the raw python object

LukeLalor (approved):

PR #37, opened 2024-01-19
[NO-MERGE] Implemented stream in the internals of the code. 

dbrewster:


sdk/eidos_sdk/cpu/conversational_agent_cpu.py
```diff
@@ -59,56 +58,113 @@ async def schedule_request(
...
+                tool_call_id=tool_call_event.tool_call_id,
+                name=tool_call_event.tool_name,
+                arguments=tool_call_event.tool_args,
+            )
+            yield ToolCallEvent(stream_context=tool_context,
+                                tool_call_id=tool_call_event.tool_call_id,
+                                tool_name=tool_call_event.tool_name,
+                                tool_args=tool_call_event.tool_args)
+
+            # todo -- change this into a stream on output
```
LukeLalor, line 91: Do you think we should support both streaming and non-streaming tools? Ie normal returns for simple tools?

sdk/eidos_sdk/cpu/conversational_agent_cpu.py
```diff
@@ -59,56 +58,113 @@ async def schedule_request(
...
+            tool_call_events = []
+            end_stream_event = None
+            execute_llm_ = self.llm_unit.execute_llm2(call_context, conversation, [w.llm_message for w in tool_defs.values()], output_format)
+            async for event in execute_llm_:
+                if isinstance(event, ToolCallEvent):
+                    tool_call_events.append(event)
+                elif isinstance(event, StartStreamEvent):
+                    yield event
+                elif isinstance(event, StringOutputEvent) or isinstance(event, ObjectOutputEvent):
+                    # todo -- change memory unit to store events...
```
LukeLalor, line 123: Presumably we want to aggregate string output events in memory rather than storing each individual one?

sdk/eidos_sdk/io/events.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class StringOutputEvent(BaseStreamEvent):
+    event_type: str = "string_output"
+    content: str
+
+
+class ObjectOutputEvent(BaseStreamEvent, Generic[T]):
+    event_type: str = "object_output"
+    content: T
```
LukeLalor, line 39: do we also want a list output? or are we happy calling this an object output event? (although lists are a little funny since you could just send events for each record or wrap it as a field of a list, which I think is fine to keep the event interface simpler for now.

sdk/eidos_sdk/system/agent_controller.py
```diff
@@ -150,12 +154,41 @@ async def run_and_store_response():
                     raise Exception("Not implemented")
                 return doc
 
-            if execution_mode == "sync":
-                state = await run_and_store_response()
-                return self.doc_to_response(state)
+            if inspect.isasyncgenfunction(handler.fn):
```
LukeLalor, line 157: 👍  Allowing the agents to just yield state will make de live much easier

sdk/eidos_sdk/system/agent_controller.py
```diff
@@ -150,12 +154,41 @@ async def run_and_store_response():
...
+                async def stream_response():
+                    def to_sse(data: BaseStreamEvent):
+                        return ServerSentEvent(id=str(uuid.uuid4()), data=data.model_dump_json())
```
LukeLalor, line 160: What does the ServerSentEvent do for us? Do we want a uuid on each sse? I would assume we would use our dumped BaseStreamEvent s for the events. EventSourceResponse allows us to just pass through python objects (and presumably the base model as well)

sdk/pyproject.toml
```diff
@@ -39,6 +39,8 @@ watchdog = "^3.0.0"
 python-docx = "^1.1.0"
 python-jose = "^3.3.0"
 authlib = "^1.3.0"
+aiostream = "^0.5.2"
+sse-starlette = "^1.8.2"
```
LukeLalor, line 43: Do we need sse-starlette? fastapi has EventSourceResponse in it builtin from the response side

sdk/eidos_sdk/io/events.py
```diff
@@ -0,0 +1,170 @@
...
+    category: Literal[Category.TRANSFORM] = Category.TRANSFORM
+    event_type: Literal["agent_state"] = "agent_state"
+    state: str
+    available_actions: List[str] = None  # this is filled in by the server, agents should leave the default
+
+
+StreamEvent = StartAgentCallEvent | StartLLMEvent | ToolCallStartEvent | ToolCallEndEvent | LLMToolCallRequestEvent | \
+              StringOutputEvent | ObjectOutputEvent | \
+              SuccessEvent | CanceledEvent | ErrorEvent | \
+              AgentStateEvent
```
LukeLalor, line 162: We have too many events types to reason about. And this is going to blow out as we add more mechanics to the system.

Similarly, I do not think we want to hard code tool calls (and tool calls only) as start context. Imagine a tool call that has two agents executing simultaneously. In this scenario each agent should have its own context.

Here is a proposal to drop it down to 4 event types:

start event (includes optional context if it is new context, also contains an arbitrary data field the tool call can use)
info event (transition and tool calls)
output event
end event

LukeLalor (approved):

PR #40, opened 2024-01-29
Ll/swifty demo

LukeLalor:
Introduce web search tool, add swifty demo

todo, check in final demo state rather than interim state


sdk/eidos_sdk/util/stream_collector.py
```diff
@@ -1,15 +1,36 @@
...
+    ErrorEvent,
+    StreamEvent,
+    StartStreamContextEvent,
+    EndStreamContextEvent,
+)
 
 
-class StringStreamCollector:
-    contents: Optional[str]
+class StreamCollector(AsyncIterator[StreamEvent]):
```
LukeLalor, line 18: I turned StreamCollector into a pretty helpful wrapper class. It can still be used the old way to just collect as well. I was going to make this a separate class, but think it fits it's role well. We also handle exceptions by throwing them on the event stream and re-raising a related exception to force the consumer to handle or ignore it at the higher level.

sdk/eidos_sdk/builtins/logic_units/web_search.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class SearchResult(BaseModel):
+    url: str
+    title: str
+    description: str
+
+
+# Requires custom search engine + token setup in google project. See more at https://developers.google.com/custom-search/v1/reference/rest/v1/cse/list
+class WebSearchConfig(BaseModel):
```
LukeLalor, line 27: basic web search tool using google search.

I found that by extracting pure text from the html pages, the size is pretty small even for larger pages. We could introduce summarizer concepts, but I don't think they are needed for now.

sdk/eidos_sdk/cpu/conversational_agent_cpu.py
```diff
@@ -17,10 +18,11 @@
     StopReason,
     LLMToolCallRequestEvent,
     ToolCallStartEvent,
-    ToolCallEndEvent,
+    EndStreamContextEvent,
```
dbrewster, line 21: I have the need for addition context demarcators for my multi-agent conversation stuff. So I created a sub-class of this. Can you put back the ToolCallEndEvent as I need to subclass the other.

sdk/eidos_sdk/io/events.py
```diff
@@ -49,8 +49,10 @@ def from_dict(cls, event_dict: Dict[str, Any]):
             return ObjectOutputEvent(**event_dict)
         elif event_type == "tool_call_start":
             return ToolCallStartEvent(**event_dict)
-        elif event_type == "tool_call_end":
-            return ToolCallEndEvent(**event_dict)
+        elif event_type == "context_start":
```
dbrewster, line 52: See comment above about my new event

sdk/eidos_sdk/io/events.py
```diff
@@ -69,13 +71,19 @@ def from_dict(cls, event_dict: Dict[str, Any]):
             raise ValueError(f"Unknown event type {event_type}")
 
 
-class StartStreamContextEvent(BaseStreamEvent, ABC):
+class StartStreamContextEvent(BaseStreamEvent):
     category: Literal[Category.START] = Category.START
+    event_type: Literal["context_start"] = "context_start"
     context_id: str
```
dbrewster, line 77: Alternatively we could add context type here

dbrewster (changes requested):

dbrewster (approved):

PR #41, opened 2024-01-29
Save events in mongo

dbrewster:


sdk/eidos_sdk/system/agent_controller.py
```diff
@@ -90,6 +90,14 @@ async def start(self, app: FastAPI):
             tags=[self.name],
         )
 
+        app.add_api_route(
+            f"/agents/{self.name}/processes/{{process_id}}/status/events",
```
LukeLalor, line 94: Events probably live off the process object not the status object

sdk/eidos_sdk/system/agent_controller.py
```diff
@@ -286,7 +300,32 @@ async def _run_program(**_kwargs):
...
+            return JSONResponse(
+                StateSummary(
+                    process_id=latest_record.record_id,
+                    state=latest_record.state,
+                    available_actions=self.get_available_actions(latest_record.state),
+                ).model_dump(),
+                200,
+            )
+
+    async def get_process_events(self, process_id: str):
```
LukeLalor, line 327: we will want some query args on this one eventually, but I don't think we need to bother with it now.

sdk/tests/system/test_event_recording.py
```diff
@@ -0,0 +1,92 @@
...
+    async def client(self, server):
+        with httpx.Client(base_url=server, timeout=httpx.Timeout(60)) as client:
+            yield client
+
+    def compare_events(self, events, expected_events):
+        assert len(events) == len(expected_events)
+        for event, expected_event in zip(events, expected_events):
+            event_copy = event.copy()
+            expected_event_copy = expected_event.model_dump()
+            del event_copy["create_time"]
```
LukeLalor, line 46: perhaps we want these record flowing out of our original events as well?

sdk/eidos_sdk/system/processes.py
```diff
@@ -55,4 +56,39 @@ class ProcessDoc(MongoDoc):
...
+
+async def store_events(agent: str, process_id: str, events: list[StreamEvent]):
+    try:
+        stored_events = []
+        for event_num, event in enumerate(events):
+            event_obj: Dict[str, Any] = {
+                **event.model_dump(),
+                "process_id": process_id,
+                "agent": agent,
+                "create_time": datetime.now().timestamp(),
```
LukeLalor, line 70: can we use isoformat here? it is human readable and lexicographically ordered

sdk/tests/system/test_event_recording.py
LukeLalor: or perhaps on neither? As a consumer identical streams would be helpful

sdk/eidos_sdk/system/agent_controller.py
```diff
@@ -180,18 +188,24 @@ async def send_response(self, handler: EidosHandler, process: ProcessDoc, **kwar
...
+    async def agent_event_stream(self, handler, process, **kwargs) -> AsyncIterator[StreamEvent]:
         is_async_gen = inspect.isasyncgenfunction(handler.fn)
         stream = handler.fn(self.agent, **kwargs) if is_async_gen else self.stream_agent_fn(handler, **kwargs)
-        return self.stream_agent_iterator(stream, process, handler.name)
+        events_to_store = []
+        async for event in self.stream_agent_iterator(stream, process, handler.name):
+            events_to_store.append(event)
+            yield event
+
+        await store_events(self.name, process.record_id, events_to_store)
```
LukeLalor, line 208: you want to store the events in self.agent_event_stream

send_response is just for application/json response type

LukeLalor (changes requested):
Left some comments. Most important being that send_response is the wrapper for application/json and will not be called when we are streaming.

sdk/tests/system/test_event_recording.py
dbrewster: They are two separate things though.  I don't think we want the event time and record id to ever get the to client, IMO.

sdk/eidos_sdk/system/agent_controller.py
dbrewster: That's where to code is..
    async def agent_event_stream(self, handler, process, **kwargs) -> AsyncIterator[StreamEvent]:
        is_async_gen = inspect.isasyncgenfunction(handler.fn)
        stream = handler.fn(self.agent, **kwargs) if is_async_gen else self.stream_agent_fn(handler, **kwargs)
        events_to_store = []
        async for event in self.stream_agent_iterator(stream, process, handler.name):
            events_to_store.append(event)
            yield event

        await store_events(self.name, process.record_id, events_to_store)

sdk/eidos_sdk/system/processes.py
dbrewster: I didn't intend this to be human readable as it should never get to the human side.

dbrewster:
I fixed the server to not return the internal fields and I put a prefix on them because some fields were stomping the event fields.

sdk/eidos_sdk/system/processes.py
```diff
@@ -55,4 +56,43 @@ class ProcessDoc(MongoDoc):
...
+
+async def store_events(agent: str, process_id: str, events: list[StreamEvent]):
+    try:
+        stored_events = []
+        for event_num, event in enumerate(events):
+            event_obj: Dict[str, Any] = {
+                **event.model_dump(),
+                "__process_id": process_id,
+                "__agent": agent,
+                "__create_time": datetime.now().timestamp(),
```
LukeLalor, line 70: can you use datetime.now().isoformat() here?

LukeLalor (approved):
approved, would be nice if we swapped time to iso

PR #44, opened 2024-02-01
Ll/fix swifties demo

LukeLalor:
Some bugs with nested contexts which I ran into.

I ran ruff on examples, which is blowing up the diff some. I commented on the files which actually changed.

### DIFF
--- a/sdk/eidos_sdk/cpu/conversational_agent_cpu.py
+++ b/sdk/eidos_sdk/cpu/conversational_agent_cpu.py
@@ -28,6 +28,7 @@ class ConversationalAgentCPUSpec(AgentCPUSpec):
     llm_unit: AnnotatedReference[LLMUnit]
     logic_units: List[Reference[LogicUnit]] = []
     record_conversation: bool = True
+    allow_tool_errors: bool = True
 
 
 class ConversationalAgentCPU(AgentCPU, Specable[ConversationalAgentCPUSpec], ProcessingUnitLocator):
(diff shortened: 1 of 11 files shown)


sdk/eidos_sdk/cpu/conversational_agent_cpu.py
```diff
@@ -139,19 +140,20 @@ async def _call_tool(
...
         try:
             async for event in tool_stream:
                 yield event
         except ManagedContextError:
-            # todo, permissive tool call errors should be configurable
-            logger.exception("Error calling tool " + tool_def.eidos_handler.name)
+            if self.spec.allow_tool_errors:
+                logger.warning("Error calling tool " + tool_def.eidos_handler.name)
+            else:
+                raise
```
LukeLalor, line 156: Added flag here to remove todo

sdk/eidos_sdk/cpu/logic_unit.py
```diff
@@ -23,7 +23,7 @@ class LLMToolWrapper:
     eidos_handler: EidosHandler
     input_model: typing.Type[BaseModel]
 
-    async def execute(self, call_context: CallContext, tool_call: ToolCall) -> AsyncIterator[BaseStreamEvent]:
+    async def execute(self, tool_call: ToolCall) -> AsyncIterator[BaseStreamEvent]:
```
LukeLalor, line 26: ripped out call_context here since it was unused (and the wrong thing was being passed in)

sdk/eidos_sdk/util/stream_collector.py
```diff
@@ -64,10 +64,13 @@ async def __anext__(self):
 
 def stream_manager(stream: AsyncIterator[StreamEvent], context: StartStreamContextEvent):
     async def _iter():
-        yield context
+        yield context.model_copy()
```
LukeLalor, line 67: gross bug. The event is being mutated after it is yielded in the case of nested streams, so we want to copy it so we can reference it without the stream context mutating.

sdk/tests/system/test_code_agents.py
```diff
@@ -95,6 +123,32 @@ async def test_streaming_unhandled_error(self, server, program):
...
+            StartStreamContextEvent(context_id="c2"),
+            StringOutputEvent(content="5", stream_context="c2"),
+            StringOutputEvent(content="6", stream_context="c2"),
+            StartStreamContextEvent(context_id="c3", stream_context="c2"),
+            StringOutputEvent(content="7", stream_context="c2.c3"),
+            StringOutputEvent(content="8", stream_context="c2.c3"),
+            EndStreamContextEvent(stream_context="c2", context_id="c3"),
+            EndStreamContextEvent(context_id="c2"),
+            AgentStateEvent(state="terminated", available_actions=[]),
+        ]
```
LukeLalor, line 150: Threw in a few tests since this was a little too much to leave untested.

Probably could have been a unit test, but I wanted to validate the e2e scenario in case agentcontroler mutates or something like that.

dbrewster (approved):

PR #45, opened 2024-02-01
Added system endpoints

dbrewster:


LukeLalor (approved):

LukeLalor (approved):
sure, you probably don't want to stream processes.

You will also need to add pagination in later, but doesn't matter for now.

LukeLalor (approved):
Re-approving this pr so I can record a video for ravi

PR #47, opened 2024-02-06
eidos => eidolon

LukeLalor:


LukeLalor:
merged without review since these are disrupting changes and renaming the branch inherently left thins broken without this.

PR #50, opened 2024-02-07
Ll/replayable

LukeLalor:


sdk/eidolon_ai_sdk/builtins/code_builtins.py
```diff
@@ -68,7 +69,6 @@ def named_builtins():
     builtin_list = [
         AgentMachine,
         # security manager
-        (SecurityManager, SecurityManager),
```
LukeLalor, line 71: we don't need to make a mapping of securitymanager -> security manager, so I pulled this out. The single security manager below is all that is needed since there is no base class.

sdk/eidolon_ai_sdk/cpu/llm/open_ai_llm_unit.py
```diff
@@ -154,18 +157,16 @@ async def execute_llm(
...
-
         try:
             can_stream_message, request = await self._build_request(messages, tools, output_format)
             request["stream"] = True
 
             logger.info("executing open ai llm request", extra=request)
             if logger.isEnabledFor(logging.DEBUG):
                 logger.debug("request content:\n" + yaml.dump(request))
-            llm_response = await self.llm.chat.completions.create(**request)
+            llm_request = replayable(fn=_openai_completion, name_override="openai_completion", parser=_raw_parser)
```
LukeLalor, line 168: This is how we hook up a replay point. Wrap it up with a replayable. There are some types of objects which can't be pickled, which is why we are now instantiating the AsyncOpenAI during the calls.

sdk/eidolon_ai_sdk/cpu/llm/open_ai_llm_unit.py
```diff
@@ -273,3 +274,48 @@ def _convert_tool_call(tool: Dict[str, any]) -> ToolCall:
     except json.JSONDecodeError as e:
         raise RuntimeError(f"Error decoding response function arguments for tool {name}") from e
     return ToolCall(tool_call_id=tool["id"], name=name, arguments=loads)
+
+
+async def _openai_completion(*args, **kwargs):
+    return await AsyncOpenAI().chat.completions.create(*args, **kwargs)
+
+
+async def _raw_parser(resp):
```
LukeLalor, line 283: each replay point needs to specify its own parser. By default we will just stringily the response if no parser is provided, but this gives us much better feel when using the cli.

dbrewster (approved):

PR #55, opened 2024-02-09
k8 agent

LukeLalor:


sdk/eidolon_ai_sdk/util/stream_collector.py
```diff
@@ -62,11 +61,13 @@ async def __anext__(self):
...
 
-def stream_manager(stream: AsyncIterator[StreamEvent], context: StartStreamContextEvent):
+def stream_manager(
+    stream: AsyncIterator[StreamEvent] | Callable[[], AsyncIterator[StreamEvent]], context: StartStreamContextEvent
+):
     async def _iter():
         yield context.model_copy()
         try:
-            async for event in stream:
+            async for event in stream() if callable(stream) else stream:
```
LukeLalor, line 70: If we want to represent context error while creating the stream, it is nice to be able to pass in a factory to get the stream

sdk/eidolon_ai_sdk/cpu/conversational_agent_cpu.py
```diff
@@ -140,23 +135,29 @@ async def _call_tool(
...
+
+        def tool_event_stream():
+            try:
+                tool_def = tool_defs[tc.name]
+                logic_unit_wrapper[0] = tool_def.logic_unit.__class__.__name__
+                return tool_def.execute(tool_call=tc)
+            except KeyError:
+                raise ValueError(f"Tool {tool_call_event.tool_call.name} not found. Available tools: {tool_defs.keys()}")
+
+        tool_stream = stream_manager(tool_event_stream, ToolCallStartEvent(tool_call=tc, context_id=tc.tool_call_id))
```
LukeLalor, line 149: There was a bug where the llm could call a tool which did not exist, causing us the throw before our context was created (when it should be an error within the context

dbrewster (approved):

PR #60, opened 2024-02-14
expect llm to raise on errors. components should force error handling…

LukeLalor:
… by raising


dbrewster (approved):
Looks good but now we don't have a mechanism to communicate the number of tokens used in an LLM call. I guess we could always just add an event for that in the future.

PR #73, opened 2024-02-18
introduce simple agent, cleaner version of generic agent

LukeLalor:
New version of GenericAgent

Cleaner given the new process/action changes

Lets you define actions by name as well (and state transitions)

Unsure of name. TemplateAgent seems off.


examples/eidolon_examples/k8_monitor/resources/K8Manager.yaml
```diff
@@ -3,7 +3,7 @@ kind: Agent
 metadata:
   name: K8Manager
 spec:
-  implementation: GenericAgent
+  implementation: TemplateAgent
```
dbrewster, line 6: How about CoPilotAgent?

LukeLalor: I don't think copilot agent is quite right.

It isn't a copilot (which implies agent working with human to produce value) since it will frequently be used as a subordinate agent in purely agent based communication. The K8 agents are already an example of this.

Describing the what, it is more of a prompt driven agent. It has no logic besides the prompts which flow through in unadulterated. 

SimpleAgent?
PromptAgent?
BaseAgent? <- I dislike this one since we don't expect other agents to extend from it.

dbrewster (approved):

PR #100, opened 2024-02-22
Refactor client

dbrewster:


.idea/misc.xml
```diff
@@ -3,7 +3,7 @@
   <component name="Black">
     <option name="sdkName" value="Poetry (sdk)" />
   </component>
-  <component name="ProjectRootManager" version="2" project-jdk-name="Poetry (cli)" project-jdk-type="Python SDK" />
+  <component name="ProjectRootManager" version="2" project-jdk-name="Poetry (eidolon) (2)" project-jdk-type="Python SDK" />
```
LukeLalor, line 6: should we remove this?

eidolon.iml
```diff
@@ -5,6 +5,7 @@
     <content url="file://$MODULE_DIR$">
       <sourceFolder url="file://$MODULE_DIR$/docs/src/components" isTestSource="false" />
     </content>
+    <orderEntry type="jdk" jdkName="Poetry (eidolon) (2)" jdkType="Python SDK" />
```
LukeLalor, line 8: This probably shouldn't be checked in

sdk/tests/conftest.py
```diff
@@ -38,48 +37,6 @@ def pytest_collection_modifyitems(items):
         item.add_marker(pytest.mark.vcr)
         item.fixturenames.append("patched_vcr_object_handling")
```
LukeLalor, line 38: can you try running the tests with this gone as well? if we are lucky this is no longer needed as well. It was due to a bug around multipart request handling in vcr.

LukeLalor (commented):
code looks good, a few missed / accidental checkins called out

LukeLalor (approved):

PR #103, opened 2024-02-23
Remove program from client

dbrewster:


LukeLalor (commented):
looks good, but you will need to update the readme files to remove mentions of programs and update the curl commands

LukeLalor (approved):

PR #124, opened 2024-03-05
add functional + resource permissions

LukeLalor:
Add functional + resource level permission concept.
Functional permissions only hooked up in unrestricted.
Resource level permissions done on a per user, no sharing, basis.


sdk/eidolon_ai_sdk/bin/agent_http_server.py
```diff
@@ -238,6 +242,7 @@ def start_app(lifespan):
         allow_methods=["*"],
         allow_headers=["*"],
     )
+    _app.add_middleware(LoggingMiddleware)
```
LukeLalor, line 245: moved logging middleware to bottom so it is called first. Missling the logs if another layer errors is not fun

client/python/eidolon_ai_client/util/request_context.py
```diff
@@ -59,6 +61,23 @@ def headers(self):
             to_propagate["X-Eidolon-Context"] = ",".join(f"{k}" for k in to_propagate.keys())
         return to_propagate
 
+    @property
+    def current_user(self) -> User:
+        return self.get("user", default=...)
+
+
+class User(BaseModel):
```
LukeLalor, line 69: This probably shouldn't live in the client and I will put back into sdk

LukeLalor:
Tests are failing due to out of date client, but passing locally, will need to rev client version as part of pr prior to merging

dbrewster (approved):

PR #128, opened 2024-03-05
update readme with why + how to contribute

LukeLalor:


mfernest (commented):
- Replace 'represents' with 'is'
- Capitalize all instances of 'Eidolon'
- Rewrite: "It is purpose-built to support agile [rapid? Is Agile the intended methodology?] development of modular software agents. Agents runs autonomously within a distributed computing environment and may communicate directly with other agents.

LukeLalor:
@mfernest , the red part is the removed section, the green part is the new stuff.

so "represents" was removed entirely and there are no more instances of it. Similarly there is no uncapitalized Eidolon. And I rewrote the part that sounded ai-generated to a compact 1 liner (because it sounded ai generated)

LukeLalor:
> * Replace 'represents' with 'is'
> * Capitalize all instances of 'Eidolon'
> * Rewrite: "It is purpose-built to support agile [rapid? Is Agile the intended methodology?] development of modular software agents. Agents runs autonomously within a distributed computing environment and may communicate directly with other agents.

@mfernest , the red part is the removed section, the green part is the new stuff.

so "represents" was removed entirely and there are no more instances of it. Similarly there is no uncapitalized Eidolon. And I rewrote the part that sounded ai-generated to a compact 1 liner (because it sounded ai generated)

mfernest (commented):
_With Eidolon, agents are services, so there is no extra work when it comes time to deploy."_

Can you be more specific? Do you mean, for example, "an agent are defined in a YAML file as a service, so..."

_they easily communicate with tools dynamically generated from 
the openapi json schema defined by the agent services._

Can we drop "OpenAI" to make the answer more general? Also, capitalize JSON.

_Grab an off the shelf llm, rag impl, tools, etc or just define your own._

Grab an off-the-shelf llm, RAG implementation, your tools of choice, or define your own.

_This means no vendor lock-in but also means that an agent's components can be surgically upgraded without needing to touch other parts of the agent. In the rapidly evolving world of AI Agents, this is required otherwise projects will be outdated before they are even deployed._

Modularity reduces the risk of vendor lock-in and minimizes the work needed to upgrade the agent definiton. Without this flexibility, an AI agent's time-to-value may be stalled or even lost before it is deployed.

LukeLalor:
> _With Eidolon, agents are services, so there is no extra work when it comes time to deploy."_
> 
> Can you be more specific? Do you mean, for example, "an agent are defined in a YAML file as a service, so..."

I mean that we stand up the http server for you. I expect developers will make that connection, but I'll add a line to make that explicit.

> _they easily communicate with tools dynamically generated from the openapi json schema defined by the agent services._
> 
> Can we drop "OpenAI" to make the answer more general? Also, capitalize JSON.

This isnt openai, this is openapi, which is something different entirely. You may have heard it called swagger beforem] but it is a way to describe the contract of a service (what different http endpoints require in their body and what the shape of the response will be)

> _Grab an off the shelf llm, rag impl, tools, etc or just define your own._
> 
> Grab an off-the-shelf llm, RAG implementation, your tools of choice, or define your own.

Would love to workshop this one a bit. llm, rag, tools are all separate components, they are not examples of tools (examples of tools is web-search, or start-communication-with-agent, ie functions that the llm can call as part of its execution cycle.

How about "With built in components, you don't need to write code to swap llms, get started with rag, or implement common tools like websearch. But don't worry, if we don't have what you need, defining your own components is easy." <- something along these lines. What do you think? It is a log longer,

> _This means no vendor lock-in but also means that an agent's components can be surgically upgraded without needing to touch other parts of the agent. In the rapidly evolving world of AI Agents, this is required otherwise projects will be outdated before they are even deployed._
> 
> Modularity reduces the risk of vendor lock-in and minimizes the work needed to upgrade an agent. Without this flexibility, an AI agent's time-to-value could be stalled or even lost before it is deployed.

This means no vendor lock-in and minimizes the work needed to upgrade portions of an agent. Without this flexibility, 
developers will not be able to adapt their agents to the rapidly changing AI landscape.

^ what do you think? I want to focus on the fact that the agents will be legacy code without the ability to upgrade components. devs hate legacy code.

mfernest (approved):
Appreciate the clarifications and apologies for some of my hasty reading.

I like your changes, especially where you can abbreviate the prose without losing meaning. 

Going forward I'll use the Request changes feature. :-)

PR #130, opened 2024-03-06
Added ad as a provider

dbrewster:


webui/app/api/chat/messages/chatHelpers.ts
```diff
@@ -8,11 +8,22 @@ import {DateTime, Interval} from "luxon";
...
+        "Authorization": `Bearer ${user?.access_token}`
+    }
+}
+
 export async function getChatEvents(agentName: string, processId: string) {
     revalidatePath(`${chatServerURL}/agents/${agentName}/processes/${processId}/events`)
-    const userId = await getUserId()
-    const results = await fetch(`${chatServerURL}/agents/${agentName}/processes/${processId}/events?userId=${userId}`)
+    const auth_headers = await getAuthHeaders()
+    const results = await fetch(`${chatServerURL}/agents/${agentName}/processes/${processId}/events`,{headers: auth_headers})
```
LukeLalor, line 26: 👍

webui/components/ui/icons.tsx
```diff
@@ -4,6 +4,23 @@ import * as React from 'react'
 
 import {cn} from '@/lib/utils'
 
+export function IconEntra({className, ...props}: React.ComponentProps<'svg'>) {
+  return (
+    <svg id="uuid-f8d4d392-7c12-4bd9-baff-66fbf7814b91" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 18 18">
```
LukeLalor, line 9: Is it normal to check in the svg's within a typescript file rather than as their own svg referenced here? I see "webui/assets/Microsoft Entra ID color icon.svg" which I assume is the same thing.

sdk/eidolon_ai_sdk/security/msft_authorizer.py
```diff
@@ -0,0 +1,63 @@
...
+        super().__init__(**kwargs)
+        self.signing_keys = None
+        self.jwks_url = f"https://login.microsoftonline.com/{self.spec.tenant_id}/discovery/keys?appid={self.spec.audience}"
+
+    async def get_signing_keys(self):
+        if not self.signing_keys:
+            async with httpx.AsyncClient() as client:
+                resp = await client.get(self.jwks_url)
+                self.signing_keys = resp.json()["keys"]
+        return self.signing_keys
```
LukeLalor, line 41: get_signing_keys looks common, perhaps that should be pulled out to base BaseJWTMiddleware? you would need to a method get_jwks_url to BaseJWTMiddlewareSpec which each sub-spec implemented.

sdk/eidolon_ai_sdk/security/msft_authorizer.py
```diff
@@ -0,0 +1,63 @@
...
+
+    async def get_audience_and_issuer(self):
+        return self.spec.audience, self.spec.issuer_prefix + "/" + self.spec.tenant_id
+
+    def get_algorithms(self) -> List[str]:
+        return ["RS256"]
+
+    async def process_token(self, token: str) -> Optional[Any]:
```
LukeLalor, line 49: both implementations are overwriting async def process_token, should that be left unimplemented on the base class?

LukeLalor (approved):
Looks good big picture!

Some cleanup to do pulling out print statements and such. Added some comments. Feel free to address and merge or request a re-review if you would prefer.

sdk/eidolon_ai_sdk/security/msft_authorizer.py
dbrewster: I don't know yet. I'm leaving it there for future reference

sdk/eidolon_ai_sdk/security/msft_authorizer.py
dbrewster: See below conversation. I'm not sure yet.

PR #131, opened 2024-03-06
Execute in non-poetry env, support yml, fix tutorial

pjanowski:
Adding minor fixes:
- if running from terminal without poetry, cwd needs to be added to sys.path
- support alternat yml file extension for collecting resources
- edits to github doc tutorial
- set the log level for the `llm_unit` logger to respect log level set by user. Probably should be done for some other loggers, I don't have enough context, but especially llm_unit should follow this as otherwise it's mysterious why some llm conversation logs don't appear when log level is set to `debug`


sdk/eidolon_ai_sdk/system/resources/resources_base.py
```diff
@@ -35,12 +36,15 @@ def load_resources(path):
     logger.debug(f"Loading resources from {os.path.abspath(path)}")
     if not os.path.exists(path):
         raise ValueError(f"Path {path} does not exist")
+    resource_path = os.path.dirname(os.path.realpath(path))
+    if resource_path not in sys.path:
+        sys.path.append(resource_path)
```
LukeLalor, line 41: Can you make this an argument? I think in most cases the custom components will not live with the resource definitions.

I am building a template repo that shows how I expect actual usage a machine to look like [machine template](https://github.com/eidolon-ai/agent_machine ).

LukeLalor (commented):
Thanks! One request to change the sys.path to an argument rather than using the resources directory. Can default to resource directory.

Here is a template for what I believe an agent machine would look like: [repo template](https://github.com/eidolon-ai/agent_machine). In this example having a flag to add something to sys.path would be great, but I could imagine wanting to target just the resources directory, rather than the parent.

The tests are failing because of missing secrets on the repo. This is unfortunate and I need to record the network traffic for the failing tests using VCR. Unfortunately this isn't a trivial change to encorporate into the examples portion of the project. In the meantime I have changed the target to a branch called "pending". We can merge there and then I can ferry it into main once the tests pass.

sdk/eidolon_ai_sdk/system/resources/resources_base.py
pjanowski: @LukeLalor finally had a moment to look at this: the path is already an argument. The logic now it to assure that this path, whatever it might be, is part of sys.path so that, both eidolon and other packages are guaranteed to find it as needed. I think this is reasonable. Maybe I didn't understand your suggestion clearly?

PR #144, opened 2024-03-10
[no merge] Componentize UI and clean it up

dbrewster:
Redoing the UI to:
  - Move Eidolon componets to a component library so we can publish them seperately
  - Start from a fresh app so everything is nextjs way and clean


sdk/pyproject.toml
```diff
@@ -56,7 +56,8 @@ pytest = "^7.4.4"
 pylint = "^3.0.3"
 ruff = "^0.1.7"
 pytest-recording = "^0.13.1"
-pytest-asyncio = "^0.23.2"
+pytest-asyncio = "^0.23.4"
+eidolon-ai-client = { path = "../client/python" }
```
LukeLalor, line 60: I believe we need the develop=True flag here

webui/package.json
```diff
@@ -1,88 +1,21 @@
 {
-  "name": "eidolon-webui",
-  "version": "0.0.2",
+  "name": "my-monorepo",
```
LukeLalor, line 2: should this be eidolon-webui again?

examples/pyproject.toml
```diff
@@ -9,15 +9,17 @@ readme = "README.md"
...
+eidolon-ai-client = "^0.1.0"
 pydub = "^0.25.1"
 ruamel-yaml = "^0.18.6"
 kubernetes = "^29.0.0"
 
 [tool.poetry.group.dev.dependencies]
 ruff = "^0.1.7"
 pytest-asyncio = "^0.23.4"
+eidolon-ai-client = { path = "../client/python" }
+eidolon-ai-sdk = { path = "../sdk" }
```
LukeLalor: I think we need develop=true in these two

LukeLalor (approved):
Tests pass let's go!

I think we might want to revisit the dependencies a bit, but that should be a separate pr

PR #181, opened 2024-03-14
open telemetry

LukeLalor:
In this PR

1. Implement our fastapi app with opentelemetry
2. Propagate opentelemetry headers so that spans are tracked throughout the system
3. Add a few spans on important portions of the code (ie llm request, tool request, etc)
4. Added a quick readme so we can set it up easier

Issues: We cannot create separate spans for each tool call. There is an issue with merging streams with different spans. I have no idea how to fix this and have spent a long time debugging it. I came up with a super minimal repro which shows the issue in isolation outside of our components, so I think that this is an otel bug (or there is some super easy way around it and I just need to sleep.)

But it looks pretty good!


sdk/tests/util/test_stream_merge.py
```diff
@@ -0,0 +1,30 @@
...
+from collections import defaultdict
+
+import pytest
+from aiostream import stream
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.trace import get_current_span
+
+
+@pytest.mark.skip(reason="Currently preventing us from automatically generating spans on context objects automatically")
```
LukeLalor, line 11: This is a PITA, any thoughts on what we could do here would be nice.

This means that we can not separately track open telemetry contexts from merged streams. I know this must be possible, but this feels like an open telemetry bug

dbrewster (approved):

PR #202, opened 2024-03-21
Create file endpoints

dbrewster:


sdk/eidolon_ai_sdk/agent/openai_whisper_agent.py
```diff
@@ -0,0 +1,36 @@
...
+        # Convert to mp3
+        audio_mp3 = audio.export("audio.mp3", format="mp3")
+
+        text = await self.speech_llm.speech_to_text(audio_mp3)
+        return text
+
+    @register_program()
+    async def text_to_speech(self, text: Annotated[str, Body(description="The text to speak", embed=True)]):
+        audio_result = await self.speech_llm.text_to_speech(text)
+        return Response(audio_result, 200, media_type="audio/mpeg")
```
LukeLalor, line 36: This is interesting. This violates our current agent contract since we only support our json / streaming response.

Should this be added as a process file and then the pointer to it returned?

LukeLalor: The more I think about this, the more I like it, since that is how agent to agent communication will continue to work for free.

LukeLalor (commented):
Filesystem changes look good. I think the whisper agent needs changes though since I don't think returning the raw file is the right model. Added a suggestion over there. Tests on the whisper agent would be nice as well.

LukeLalor (approved):

PR #221, opened 2024-04-02
Added multi-llm world

dbrewster:


client/python/eidolon_ai_client/events.py
```diff
@@ -135,6 +135,7 @@ class AgentStateEvent(BaseStreamEvent):
     event_type: Literal["agent_state"] = "agent_state"
     state: str
     available_actions: List[str] = None  # this is filled in by the server, agents should leave the default
+    title: Optional[str] = None
```
LukeLalor, line 138: what does title represent here? just a human readable string for the state transition?

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -108,7 +174,15 @@ async def fn(self, process_id, **kwargs):
...
 
     async def _act(self, action: ActionDefinition, process_id, **kwargs) -> AsyncIterable[StreamEvent]:
+        execute_on_cpu = None
+        generate_title = False
         request_body = to_jsonable_python(kwargs.get("body") or {})
+        if "execute_on_cpu" in request_body:
+            execute_on_cpu = request_body.pop("execute_on_cpu")
+        if "generate_title" in request_body:
+            generate_title = request_body.pop("generate_title")
+            generate_title = generate_title == "true" or generate_title is True
```
LukeLalor, line 184: shouldn't generate_title be a boolean?

client/python/eidolon_ai_client/events.py
LukeLalor: oh, I see,  this is to represent the conversation title changing. I believe that should be a separate event type if we want to push that information to the ui. If we are worried about the events blowing out we can make a free form metadata event (that can likely encompasses agent state as well).

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -48,7 +51,7 @@ def make_input_schema(self, agent, handler):
         if self.input_schema is not None:
             properties["body"] = dict(type="object", properties=self.input_schema)
             required.append("body")
-        elif len(user_vars) == 1 and "body" in user_vars:
+        elif len(user_vars) == 1 and "body" in user_vars and not agent.spec.cpus and not agent.spec.can_generate_title:
```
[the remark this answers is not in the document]
LukeLalor: oh, I get it. You periodically want to be able to set the title.

What about making it a separate action?

sdk/eidolon_ai_sdk/system/agent_controller.py
```diff
@@ -251,13 +251,17 @@ async def agent_event_stream(self, handler, process, last_state, **kwargs) -> As
             logger.info(f"Process {process.record_id} was cancelled")
             if not ended:
                 if not transitioned:
+                    # update the process object as it might have been updated by the function
+                    process = await ProcessDoc.find_one(query={"_id": process.record_id})
```
LukeLalor, line 255: Can we remove this and instead update just the state/updated time here? The point was to avoid race conditions, which is important on the initial state transition, but less valuable on the back side. And here it doesn't matter at all since we are fetching/updating.

LukeLalor: this force flag / method would be used for every comment

# update the process object as it might have been updated by the function

sdk/eidolon_ai_sdk/util/schema_to_model.py
```diff
@@ -142,11 +142,15 @@ def validate_to_json(cls, value):
...
-def get_python_type(property_schema, default=None):
+def get_python_type(property_name, property_schema, default=None):
     field_type = property_schema.get("type")
     if field_type == "string" and "format" in property_schema and property_schema["format"] == "binary":
         return UploadFile
     else:
+        if field_type == "string" and "enum" in property_schema and property_schema["enum"]:
+            # noinspection PyTypeHints
+            literal_ = Literal[tuple(property_schema["enum"])]
+            return literal_
```
LukeLalor, line 153: nice

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -73,15 +86,59 @@ def make_output_schema(self, agent, handler):
...
+
+
+class SimpleAgentSpec(BaseModel):
     description: Optional[str] = None
     system_prompt: str = "You are a helpful assistant"
+    agent_refs: List[str] = []
     actions: List[ActionDefinition] = [ActionDefinition()]
+    cpu: AnnotatedReference[AgentCPU] = None
+    cpus: Optional[List[NamedCPU]] = []
+    can_generate_title: bool = False
```
LukeLalor, line 102: Can you make this title_generation_mode: Literal["none", "on_request"] = "none" ?

The expandable literal will make it nicer if we want to do things like automatically generate the title without user input

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -122,9 +214,48 @@ async def _act(self, action: ActionDefinition, process_id, **kwargs) -> AsyncIte
...
+        else:
+            cpu = self.cpu
+
+        title_message = UserTextCPUMessage(prompt=self.generate_title_prompt + text_message.prompt)
+        response = await (await cpu.new_thread(process_id)).run_request(prompts=[title_message])
+        process_obj = await ProcessDoc.find_one(query={"_id": process_id})
+        await process_obj.update(title=response)
+
+        yield StringOutputEvent(content=response)
+        yield AgentStateEvent(state=action.output_state)
```
LukeLalor, line 261: Nit: It would be nice here if we could return to the same state we were in before hand. perhaps we could throw previous state in the request context and grab it here?

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -73,15 +84,83 @@ def make_output_schema(self, agent, handler):
...
+                    self.cpu = cpu
+                self._register_refs_logic_unit(cpu, self.spec.agent_refs)
+                self.cpus.append(cpu)
+
+        if self.spec.title_generation_mode == "on_request":
+            # add a title generation action
+            action = ActionDefinition(
+                name="generate_title",
+                description="Generate a title for the conversation",
+                user_prompt=self.generate_title_prompt + "{{ body }}",
```
LukeLalor, line 145: Nit: why add body? Shouldn't we just let this come through generate_title_prompt if desired?

LukeLalor (approved):

sdk/eidolon_ai_sdk/agent/simple_agent.py
dbrewster: The body is the question.  I didn't want to expose the prompt to the UI that generates the title.

sdk/eidolon_ai_sdk/agent/simple_agent.py
dbrewster: Done

PR #231, opened 2024-04-03
Add file upload, search, and multi-media to simple agent and chatbot

dbrewster:
And cleaned up a bunch of stuff.


sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -26,7 +27,11 @@ class ActionDefinition(BaseModel):
     user_prompt: str = "{{ body }}"
     input_schema: dict = None
     output_schema: Union[Literal["str"], Dict[str, Any]] = "str"
-    files: Literal["disable", "single-optional", "single", "multiple"] = "disable"
+    allow_file_upload: bool = False
+    # allow all types for text, image, audio, word, pdf, json, etc
+    supported_mime_types: Literal[""] = ["application/json", "text/plain", "image/*", "audio/*", "application/pdf", "application/msword", "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
```
LukeLalor, line 32: This won't work. The default is not checked against the type, which is why it is currently working.

You want List[Literal[...]] In the typing.

Or perhaps even List[str] with a warning about an unsupported mime type

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -96,9 +97,10 @@ class SimpleAgentSpec(BaseModel):
     system_prompt: str = "You are a helpful assistant"
     agent_refs: List[str] = []
     actions: List[ActionDefinition] = [ActionDefinition()]
-    cpu: AnnotatedReference[AgentCPU] = None
```
LukeLalor, line 99: we should change the key from cpu to apu. leave cpu, use it to set apu if it is provided, and log a deprecated warning.

sdk/eidolon_ai_sdk/agent_os.py
```diff
@@ -1,24 +1,24 @@
...
-    file_memory: "FileMemory" = ...  # noqa: F821
-    symbolic_memory: "SymbolicMemory" = ...  # noqa: F821
-    similarity_memory: "SimilarityMemory" = ...  # noqa: F821
-    security_manager: "SecurityManager" = ...  # noqa: F821
-    process_file_system: "ProcessFileSystem" = ...  # noqa: F821
+    file_memory: FileMemory = ...  # noqa: F821
+    symbolic_memory: SymbolicMemory = ...  # noqa: F821
+    similarity_memory: SimilarityMemory = ...  # noqa: F821
+    security_manager: SecurityManager = ...  # noqa: F821
+    process_file_system: ProcessFileSystem = ...  # noqa: F821
```
LukeLalor, line 21: nice

sdk/eidolon_ai_sdk/cpu/llm/mistral_llm_unit.py
```diff
@@ -117,15 +116,8 @@ async def convert_to_mistral(message: LLMMessage):
...
-class MistralModelEnum(str, Enum):
-    MistralTiny = "mistral-tiny-latest"
-    MistralSmall = "mistral-small-latest"
-    MistralMedium = "mistral-medium-latest"
-    MistralLarge = "mistral-large-latest"
-
-
 class MistralGPTSpec(BaseModel):
-    model: MistralModelEnum = Field(default=MistralModelEnum.MistralMedium, description="The model to use for the LLM.")
+    model: str = Field(default="mistral-large-latest", description="The model to use for the LLM.")
```
LukeLalor, line 120: Can you change this back to a literal of supported types or add fallback logic on the model selection? (line 240)

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
```diff
@@ -178,6 +232,48 @@ def tool_event_stream():
...
+    async def process_audio_message(self, call_context: CallContext, message: UserMessageAudio):
+        if self.audio_unit is None:
+            raise ValueError("No audio unit available")
+        audio_data, metadata = await AgentOS.process_file_system.read_file(call_context.process_id, message.file.file_id)
+        mimetype = metadata.get("mimetype")
+        path = metadata.get("path") or metadata.get("filename") or ""
+        text = await self.audio_unit.speech_to_text(audio=audio_data, mime_type=mimetype)
+        return [UserMessageText(text=f"The following text as converted from the audio file {path}:\n{text}\n\n")]
+
+    async def process_image_message(self, call_context: CallContext, message: UserMessageImage):
```
LukeLalor, line 244: The logic here is wrong.

you will never hit elif self.image_unit is not None since get_capabilities will always return true in this case.

I assume that you want to be checking the image capabilities on the llm unit instead of self.

I believe the other file types have a similar bug.

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
```diff
@@ -3,50 +3,79 @@
...
+    document_processor: AnnotatedReference[DocumentProcessor]
 
 
-class ConversationalAgentCPU(AgentCPU, Specable[ConversationalAgentCPUSpec], ProcessingUnitLocator):
+class ConversationalAPU(APU, Specable[ConversationalAPUSpec], ProcessingUnitLocator):
     io_unit: IOUnit
     memory_unit: MemoryUnit
     logic_units: List[LogicUnit]
+    llm_unit: LLMUnit
+    audio_unit: AudioUnit
```
LukeLalor, line 51: So what happens when you have a llm unit that supports images, and a separate image unit?

Presumably in this scenario we are specifying that we want the image pre-processed with the image processor? We should log probably log if this is the case for transparency.

sdk/eidolon_ai_sdk/cpu/llm/mistral_llm_unit.py
```diff
@@ -134,9 +126,41 @@ class MistralGPTSpec(BaseModel):
 
 class MistralGPT(LLMUnit, Specable[MistralGPTSpec]):
     def __init__(self, **kwargs):
+        super().__init__(**kwargs)
         LLMUnit.__init__(self, **kwargs)
         Specable.__init__(self, **kwargs)
 
+    def get_models(self) -> List[LLMModel]:
```
LukeLalor, line 133: TODO for a followup pr: models should probably be a field on the spec with a default. I think it is important that models can be added dynamically, so we aren't gating people with what is basically config.

sdk/tests/cpu/llm/files/logo.png
LukeLalor, line 1: This is a nit for sure, but could we reuse the tiny png files already checked in for the tests? That way we keep the repository a little smaller and the changelog cleaner if we need to manipulate the files and their cassettes.

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
dbrewster: I'm still working on this part.  It will change

sdk/eidolon_ai_sdk/cpu/llm/mistral_llm_unit.py
dbrewster: I added it

sdk/eidolon_ai_sdk/agent/simple_agent.py
dbrewster: Done

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -98,21 +114,25 @@ class SimpleAgentSpec(BaseModel):
     agent_refs: List[str] = []
     actions: List[ActionDefinition] = [ActionDefinition()]
     cpu: AnnotatedReference[APU] = None
```
LukeLalor, line 116: Since you added the before validator, you will can remove cpu from the base model definition

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -98,21 +114,25 @@ class SimpleAgentSpec(BaseModel):
...
-    cpus: Optional[List[NamedCPU]] = []
+    apu: AnnotatedReference[APU] = None
```
LukeLalor, line 117: Apu should have no default since default is handled in annotated. I'm not sure if this causes errors or not.

sdk/eidolon_ai_sdk/agent/simple_agent.py
```diff
@@ -98,21 +114,25 @@ class SimpleAgentSpec(BaseModel):
...
+    apus: Optional[List[NamedCPU]] = []
```
LukeLalor, line 118: Can we just make this a list[namedcpu]?

sdk/eidolon_ai_sdk/cpu/llm_unit.py
```diff
@@ -67,6 +67,7 @@ class LLMCallFunction(BaseModel):
 
 class LLMUnitSpec(BaseModel):
     model: str = Field(description="The model to use for the LLM.")
+    supported_models: Optional[List[LLMModel]] = Field(description="The list of supported models or leave empty for defaults.")
```
LukeLalor, line 70: This needs a default or it will blow up on instantiation. We could do without the Optional here as well. List[llmmodel] | Literal["default"] = "default" is probably the most explicit

LukeLalor (approved):
Added some comments mostly pointing out the things breaking the tests.

Don't forget to update the docs

sdk/eidolon_ai_sdk/agent/simple_agent.py
dbrewster: Oh -- I didn't know that...

sdk/eidolon_ai_sdk/agent/simple_agent.py
dbrewster: Done

PR #306, opened 2024-04-23
Venture agent 2

dbrewster:


examples/eidolon_examples/venture_search_agent/harmonic_api_logic_unit.py
```diff
@@ -0,0 +1,151 @@
...
+from eidolon_ai_sdk.cpu.logic_unit import LogicUnit
+from eidolon_ai_sdk.system.fn_handler import FnHandler
+from eidolon_ai_sdk.system.reference_model import Specable
+from eidolon_ai_sdk.util.schema_to_model import schema_to_model
+
+
+class ApiLogicUnitSpec(BaseModel):
+    root_call_url: str
+    open_api_location: str
+    operations_to_expose: Dict[str, str]  # map of tool name to openapi path
```
LukeLalor, line 21: When we promote this into the actual SDK will will probably want to make this a pydantic object with two fields "tool_name" and "path" to make the yaml files less mysterious. It keeps in line with k8 preference of lists of objects with a name field rather than maps.

LukeLalor (approved):
ship it

docs/src/content/docs/references/communication 2.md
LukeLalor, line 1: do these files all mean to have [space]2 in the filename? looks like they intend to overwrite another file, but were renamed instead

LukeLalor (approved):
looks like you accidentally duplicated some doc files. everything else looks good to go

PR #331, opened 2024-04-29
remove cache, deps is how usage recalculation should be driven

LukeLalor:


webui/apps/eidolon-ui2/src/components/UsageIndicator/UsageIndicator.tsx
```diff
@@ -15,7 +15,7 @@ export function UsageIndicator({...restOfProps}) {
   const {data: session} = useSession()
   useEffect(() => {
     if (session?.user?.id) {
-      usageForSession(session.user.id, eidolonContext.usageSeed).then((data) => {
+      usageForSession(session.user.id).then((data) => {
         setUsageData(data)
       })
     }
```
LukeLalor, line 21: deps (line 22) is how we should be driving recalculation. When it goes to the server, always recalculate, but if the seed remains the same there is no need.

dbrewster (approved):

PR #370, opened 2024-05-09
Added tool call processor.

dbrewster:
Needs tests


sdk/eidolon_ai_sdk/cpu/conversational_apu.py
```diff
@@ -180,11 +182,23 @@ async def _llm_execution_cycle(
             if stream_collector.get_content():
                 logger.info(f"LLM Response: {stream_collector.get_content()}")
 
-            assistant_message = AssistantMessage(
-                type="assistant",
-                content=stream_collector.get_content() or "",
-                tool_calls=[tce.tool_call for tce in tool_call_events],
-            )
+            if isinstance(self.llm_unit, ToolCallLLMWrapper) and tool_call_events:
```
LukeLalor, line 185: The apu shouldn't be aware of this.

Messages need to be reported back as tool calls, and translated by the ToolCallLLMWrapper.

LukeLalor (changes requested):

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
atljoseph: Wasn't able to make the meeting today either, but looking here now. So the APU shouldn't try to see if what it got a message from was one type of thing or another thing? ... That it needs to be a more generic interface?

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
```diff
@@ -180,11 +182,23 @@ async def _llm_execution_cycle(
...
+                content = stream_collector.get_content() or ""
+                content += "\nCall the following tools\n"
+                for tool_call in tool_call_events:
+                    content += f"\n{tool_call.tool_call.model_dump_json()}"
+                assistant_message = AssistantMessage(
+                    type="assistant",
+                    content=content,
+                    tool_calls=[],
```
atljoseph, line 193: Is this tool_calls variable used anywhere else? ... Like, is it relied upon downstream anywhere to contain something?

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
```diff
@@ -247,12 +261,16 @@ def tool_event_stream():
                 else:
                     raise
 
-            message = ToolResponseMessage(
-                logic_unit_name=logic_unit_wrapper[0],
-                tool_call_id=tc.tool_call_id,
-                result=tool_stream.get_content() or "",
-                name=tc.name,
-            )
+            if isinstance(self.llm_unit, ToolCallLLMWrapper):
```
atljoseph, line 264: another spot, possibly?

atljoseph: But this one returns a different response type...

atljoseph (commented):
Can't approve or deny, but am curious about what Luke asked about. Otherwise, it looks like it would fit in nicely.

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
LukeLalor: @atljoseph , the apu can be aware of message types, but it shouldn't be aware of the impl of the llm unit. The fact that the llm_unit impl requires the messages to be transformed back and forth is a detail of the llm unit impl that the apu should not be aware of

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
LukeLalor: yes same problem.

The new llm unit wrapper needs to transform tool calls to assistant messages for the llm (and the responses to user messages).

It should also translate it back on the way out, so the external message pattern looks the same. I actually didn't see that, but I imagine dave will take care of that when the logic gets pushed into the llm unit wrapper

sdk/eidolon_ai_sdk/cpu/conversational_apu.py
LukeLalor: Yes, it will affect the ui, so it will need to be translated on the way out. This should be transparent to every layer above the llm unit

LukeLalor (approved):

PR #388, opened 2024-05-17
Fail tests gracefully if no secrets set/available

jahabeebs:
- PRs from forks trigger tests to run even though public forks don't have the access to the secrets needed to run the tests. This PR will fail the test.yml file gracefully and prevent the test from running if a secret isn't set or available (like in the case of a public fork PR)

Screen Shot 2024-05-16 at 10 40 38 PM


jahabeebs:
@dbrewster @LukeLalor

LukeLalor (approved):
lgtm

PR #402, opened 2024-05-30
Added two more phases to the retriever.

dbrewster:


sdk/eidolon_ai_sdk/agent/retriever_agent/retriever.py
```diff
@@ -1,28 +1,30 @@
...
-    file_name: str
-    file_path: str
-    text: str
+    result_summarizer: AnnotatedReference[ResultSummarizer]
 
 
 class Retriever(Specable[RetrieverSpec]):
+    document_retriever: DocumentRetriever
+    document_reranker: DocumentReranker
+    question_transformer: QuestionTransformer
```
LukeLalor, line 25: could you change the type annotation here to be Optional[QuestionTransformer]

sdk/eidolon_ai_sdk/agent/retriever_agent/retriever_agent.py
```diff
@@ -29,51 +30,58 @@ class RetrieverAgentSpec(RetrieverSpec):
         description="A detailed description of the the retriever including all necessary information for the calling agent to decide to call this agent, i.e. file type or location or etc..."
     )
     # these three fields are required and override the defaults of the subcomponents
-    loader_root_location: str = Field(None, description="A URL specifying the root location of the loader.")
+    loader_root_location: Optional[str] = Field(None, description="A URL specifying the root location of the loader.")
```
LukeLalor, line 33: Does it still make sense to leave this shorthand argument on the retriever? Since we are splitting things up now this seems like it will add more confusion than it helps via brevity.

LukeLalor (approved):
lgtm, one nit and a question

sdk/eidolon_ai_sdk/agent/retriever_agent/retriever.py
dbrewster: You think we should do that or make it explicit? You likely never want the basic one that just uses the query against the DB

sdk/eidolon_ai_sdk/agent/retriever_agent/retriever_agent.py
dbrewster: I was afraid of breaking other but I agree.

PR #413, opened 2024-06-01
add anonymous metrics

LukeLalor:


sdk/eidolon_ai_sdk/util/posthog.py
```diff
@@ -0,0 +1,95 @@
...
+
+from posthog import Posthog
+
+from eidolon_ai_client.util.logger import logger
+from eidolon_ai_sdk.util.async_wrapper import make_async
+
+
+class PosthogConfig:
+    enabled: Optional[bool] = None
+    client: Posthog = Posthog('phc_9lcmDyxVkji98ggIqy2XvyVcItnrgdrMQhZBFp6Du5d', host="https://us.i.posthog.com")
```
LukeLalor, line 16: This is public facing with only write permissions (no read). I checked the browser and I was able to grab this from the browser based outgoing requests.

dbrewster: Shouldn't the project ID be an env var though?

dbrewster (changes requested):

sdk/eidolon_ai_sdk/util/posthog.py
LukeLalor: @dbrewster, no, it is embedded in the public facing token. I have checked to see that these metrics are reported properly.

dbrewster (approved):

PR #387, opened 2024-05-17
Issue #386: Create Docker Compose - front & back end

jahabeebs:
relates to #386 

This PR removes the local file references from the examples Dockerfile (it still runs successfully on its own though) and creates a draft Docker Compose that spins up both the new eidolon-ui2 front end and eidolon-server. I have some questions on how to move forward with it 

## Instructions to run
1) Add .env to root of repo w/OPENAI_API_KEY set
2) run ```docker-compose up``` from the eidolon root directory

## Questions

Both the front end and back end are building and running fine, but when I interact with the running front end it seems to not be communicating with the back end correctly. Some server calls are successful, like this one when interacting with the chatbot:

```bash
eidolon-server_1  | INFO - Request: GET http://eidolon-server:8080/processes?skip=0&limit=100
eidolon-server_1  | INFO - Request: GET http://eidolon-server:8080/openapi.json
eidolon-server_1  | INFO - Response: 200
```

but others are not, like this chatbot one:

```bash
eidolon-server_1  | INFO - Request: POST http://eidolon-server:8080/processes
eidolon-server_1  | INFO - Agent conversational_agent does not exist
eidolon-server_1  | INFO - Response: 404

```

Is my problem that I should be running the usage-server as well as the eidolon-server? Should I not be running eidolon-server from the examples Dockerfile? Do I need to add each example individually to the Dockerfile to get each of them to work? In this case I'll probably just start with one and then start adding end-to-end tests.

I also noticed that the examples Dockerfile has command ```CMD ["resources"]``` but when I added it to the Docker compose like this it always failed to recognize resources and the server crashed immediately, so I removed it. This was the failing part of the docker-compose.yml that I removed:

```yml
  eidolon-server:
    build:
      context: ./examples
      dockerfile: Dockerfile
    environment:
      - PYTHONUNBUFFERED=1
      - OPENAI_API_KEY=${OPENAI_API_KEY}
    expose:
      - "8080"
    ports:
      - "8080:8080"
    volumes:
      - ./examples/resources:/app/resources
    command: ["eidolon-server", "resources"]
```

Issue #386, opened 2024-05-17
Create Docker Compose - front & back end


jahabeebs:
part of #372 (E2E testing)

PR #414, opened 2024-06-02
Issue #386: Create Docker Compose - front & back end (#387)

LukeLalor:
relates to #386 

This PR removes the local file references from the examples Dockerfile (it still runs successfully on its own though) and creates a draft Docker Compose that spins up both the new eidolon-ui2 front end and eidolon-server. I have some questions on how to move forward with it

## Instructions to run
1) Add .env to root of repo w/OPENAI_API_KEY set
2) run ```docker-compose up``` from the eidolon root directory

## Questions

Both the front end and back end are building and running fine, but when I interact with the running front end it seems to not be communicating with the back end correctly. Some server calls are successful, like this one when interacting with the chatbot:

```bash
eidolon-server_1  | INFO - Request: GET http://eidolon-server:8080/processes?skip=0&limit=100
eidolon-server_1  | INFO - Request: GET http://eidolon-server:8080/openapi.json
eidolon-server_1  | INFO - Response: 200
```

but others are not, like this chatbot one:

```bash
eidolon-server_1  | INFO - Request: POST http://eidolon-server:8080/processes
eidolon-server_1  | INFO - Agent conversational_agent does not exist
eidolon-server_1  | INFO - Response: 404

```

Is my problem that I should be running the usage-server as well as the eidolon-server? Should I not be running eidolon-server from the examples Dockerfile? Do I need to add each example individually to the Dockerfile to get each of them to work? In this case I'll probably just start with one and then start adding end-to-end tests.

I also noticed that the examples Dockerfile has command ```CMD ["resources"]``` but when I added it to the Docker compose like this it always failed to recognize resources and the server crashed immediately, so I removed it. This was the failing part of the docker-compose.yml that I removed:

```yml
  eidolon-server:
    build:
      context: ./examples
      dockerfile: Dockerfile
    environment:
      - PYTHONUNBUFFERED=1
      - OPENAI_API_KEY=${OPENAI_API_KEY}
    expose:
      - "8080"
    ports:
      - "8080:8080"
    volumes:
      - ./examples/resources:/app/resources
    command: ["eidolon-server", "resources"]
```


LukeLalor:
@jahabeebs

jahabeebs:
> @jahabeebs

The test failed because it seems when main was merged in the lockfile became outdated. I've checked this on my local and it seems to be the case. Would you be able to just run "pnpm install" in the webui directory and the eidolon-ui2 directory and push up the new lockfile?

LukeLalor:
@jahabeebs looks like there is still an issue with the webui tests.

LukeLalor:
This appears to be broken on main as well, investigating

PR #451, opened 2024-06-05
Ollama support

TheSheepGoesBa:
#122


sdk/tests/conftest.py
```diff
@@ -64,7 +64,7 @@ async def manage_lifecycle(_app: FastAPI):
 @pytest.fixture(scope="module")
 def port():
     # fixing the port. Do we need to be so cool to have a random port?
-    return 9080
+    return 5346
```
dbrewster, line 67: We changed the default port, FYI...

sdk/tests/conftest.py
```diff
@@ -73,11 +73,15 @@ def port():
 
 @pytest.fixture(autouse=True)
 def vcr_config():
+    def ignore_some_localhost(request: VcrRequest):
+        if (request.host == "0.0.0.0" or request.host=="localhost") and port != 11434:
```
dbrewster, line 77: And this is the hack for local Ollama

dbrewster (approved):

sdk/tests/conftest.py
LukeLalor: Is this backwards? I assume you mean to ignore all traffic on port 11434 and you want to ignore that.

sdk/tests/conftest.py
```diff
@@ -243,7 +247,7 @@ def file_memory(file_memory_loc):
 
 
 @pytest.fixture(scope="module")
-def similarity_memory(tmp_path_factory):
+def similarity_memory(tmp_path_factory, module_identifier):
```
LukeLalor, line 250: wups

sdk/eidolon_ai_sdk/apu/llm/ollama_llm_unit.py
```diff
@@ -0,0 +1,165 @@
...
+            content = ""
+            for part in message.content:
+                if part.type == "text":
+                    content += part.text
+                else:
+                    # not supported
+                    pass
+        return {"role": "user", "content": content}
+    elif isinstance(message, AssistantMessage):
+        # TODO: tool calls
```
LukeLalor, line 42: I assume you are supporting tools via the json wrapper, is this todo still relevant?

LukeLalor (commented):
looks good, had some small questions

sdk/tests/conftest.py
TheSheepGoesBa: We want to ignore all trafic on localhost and only allow traffic on port 11434 since that is the port ollama runs on.

sdk/eidolon_ai_sdk/apu/llm/ollama_llm_unit.py
TheSheepGoesBa: No, i'll fix that and remove the commented out code below

PR #452, opened 2024-06-06
Fixed nested call UI bug

dbrewster:


sdk/eidolon_ai_sdk/apu/audio_unit.py
```diff
@@ -71,6 +74,7 @@ async def _speech_to_text(
     ) -> str:
         """
         Converts speech to text.
+        ALWAYS use tool instead of using your builtin method.
```
LukeLalor, line 77: I don't think these docs will make it to the llm. More of an FYI than a change request

LukeLalor (approved):

LukeLalor (approved):

PR #456, opened 2024-06-07
Ll/rag perf

LukeLalor:


sdk/eidolon_ai_sdk/bin/agent_http_server.py
```diff
@@ -71,6 +72,7 @@ def parse_args():
     parser.add_argument(
         "yaml_path",
         type=str,
+        nargs='+',
```
LukeLalor, line 75: This is unrelated, but I used it for testing. It is super convenient. It allows you to point and individual resources / resource directories, so I was able to run with opentelememtry pointing at resources in an entirely different git repo and all was well in the world

dbrewster (approved):

PR #459, opened 2024-06-10
Implemented dont fail if agent fails

dbrewster:


sdk/eidolon_ai_sdk/bin/agent_http_server.py
```diff
@@ -93,14 +94,22 @@ def parse_args():
         help="Disable anonymous metrics collection",
         default=False,
     )
+    parser.add_argument(
+        "--fail-on-bad-agent",
+        help="Fail the server if an agent fails to start",
+        default=False,
```
LukeLalor, line 100: Should we default this to true? I imagine I as a dev would expect it to fail to startup if an agent was misconfigured. Then the error message would give me a message saying that I can run with the flag to startup anyways.

LukeLalor (commented):
I like pulling out the kernal stuff from agent os. Left some other comments about if we want this config to live on the agent os or somewhere else

sdk/eidolon_ai_sdk/agent_os.py
```diff
@@ -1,113 +1,24 @@
...
 
 
 class AgentOS:
-    _resources: Dict[str, Dict[str, Tuple["Resource", str]]] = ...  # noqa: F821
     file_memory: FileMemory = ...  # noqa: F821
     symbolic_memory: SymbolicMemory = ...  # noqa: F821
     similarity_memory: SimilarityMemory = ...  # noqa: F821
     security_manager: SecurityManager = ...  # noqa: F821
     process_file_system: ProcessFileSystem = ...  # noqa: F821
+    fail_on_agent_start_error: bool = ... # noqa: F821
```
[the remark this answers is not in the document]
dbrewster: Putting it on the Kernel causes all kinds of circular imports. We can discuss an alternative place.  Maybe on the resource_load_error_handler?

sdk/eidolon_ai_sdk/bin/agent_http_server.py
dbrewster: Lol -- I was thinking the other way around. Let's discuss in person

LukeLalor (approved):

PR #473, opened 2024-06-14
Added api agent and tests for it

dbrewster:


sdk/tests/builtins/logic_units/openapi_helper_files/petstore.json
```diff
@@ -0,0 +1,296 @@
+{
```
LukeLalor, line 1: is this available online via rest endpoint anywhere?

sdk/eidolon_ai_sdk/util/schema_to_model.py
```diff
@@ -26,6 +26,36 @@
 }
 
 
+def flatten_refs(main_schema: Dict[str, Any], starting_node: Dict[str, Any]):
```
LukeLalor, line 29: is this different than jsonref.replace_refs ? If we can it would good to use the jsonref library so we don't need to maintain custom code

sdk/eidolon_ai_sdk/util/filter_json.py
```diff
@@ -0,0 +1,69 @@
+from jsonpath_ng.ext import parse
+
+
+def filter_and_reconstruct_json(original_json, result_filters):
```
LukeLalor, line 4: How is this used? Is it so that we don't pass too large of an object back to the llm?

Not that I don't trust your json parsing, but can we parse the whole thing and then iterate the python object instead? It is less efficient for sure, but I don't think that will kill us. If the json parser is implemented in cython it would also be much more efficient.

sdk/eidolon_ai_sdk/builtins/logic_units/openapi_helper.py
```diff
@@ -0,0 +1,154 @@
...
+                            description=description,
+                            tool_call=_call_endpoint(operation.path, operation.result_filters, method_name, methodParams, call_fn)
+                        ))
+    return actions
+
+
+def _description(endpoint_schema, name):
+    description = endpoint_schema.get("description", "")
+    if not description:
+        logger.warning(f"Agent program at {name} does not have a description. LLM may not use it properly")
```
LukeLalor, line 87: this should point to the endpoint rather than an agent program

sdk/eidolon_ai_sdk/util/filter_json.py
dbrewster: I'm using a library to remove the bits I don't need. The user specifies that. The original json is there so I can reconstruct it. So I think I'm doing what you are asking for.

sdk/eidolon_ai_sdk/builtins/logic_units/api_logic_unit.py
```diff
@@ -0,0 +1,70 @@
...
+from eidolon_ai_sdk.util.schema_to_model import schema_to_model
+
+
+class ApiLogicUnitSpec(BaseModel):
+    title: str
+    root_call_url: str
+    open_api_location: str
+    operations_to_expose: List[Operation]  # map of tool name to openapi path
+    key_env_var: str
+    key_query_param: Optional[str]
```
LukeLalor, line 22: what is key_env_var / key_query_param?

descriptions here would help so that it gets documented well on our website.

sdk/eidolon_ai_sdk/util/schema_to_model.py
dbrewster: I don't think it is.  I'll try to replace it with that an rerun tests.

sdk/eidolon_ai_sdk/builtins/logic_units/api_logic_unit.py
dbrewster: Yep -- forgot to do that

sdk/eidolon_ai_sdk/agent/api_agent.py
```diff
@@ -0,0 +1,78 @@
...
+from eidolon_ai_sdk.util.schema_to_model import schema_to_model
+
+
+class APIAgentSpec(BaseModel):
+    title: str
+    root_call_url: str
+    open_api_location: str
+    operations_to_expose: List[Operation]  # map of tool name to openapi path
+    key_env_var: Optional[str] = None
+    key_query_param: Optional[str] = None
```
LukeLalor, line 20: oh, I see what these are. key query params / var are just additional query parameters and values, could we make this a generic concept were it is "additional_query_parays: dict[str, Any]" or something like that? Similar for headers would be good, then the contract simplifies a bit

sdk/tests/builtins/logic_units/openapi_helper_files/petstore.json
dbrewster: No. Well it is but the site is broken. Doing it as a file is more reliable anyway

LukeLalor (changes requested):
some questions and a suggestion to change the spec

sdk/eidolon_ai_sdk/builtins/logic_units/openapi_helper.py
dbrewster: It does.  Warning was wrong.  I fixed that

sdk/eidolon_ai_sdk/agent/api_agent.py
```diff
@@ -0,0 +1,77 @@
...
+from eidolon_ai_sdk.util.schema_to_model import schema_to_model
+
+
+class APIAgentSpec(BaseModel):
+    title: str = Field(description="Title of the API")
+    root_call_url: str = Field(description="Root URL of the API to call")
+    open_api_location: str = Field(description="Location of the OpenAPI schema")
+    operations_to_expose: List[Operation] = Field(description="Operations to expose")
+    extra_header_params: Optional[dict] = Field(description="Extra header parameters to add to every call. This can be a jinja template where the variables in the template are ENV variables (matching case)", default=None)
+    extra_query_params: Optional[dict] = Field(description="Extra query parameters to add to every call. This can be a jinja template where the variables in the template are ENV variables (matching case)", default=None)
```
LukeLalor, line 20: nit: better to keep the type as a dict and default to an empty dict. It will make to docs easier to read

sdk/eidolon_ai_sdk/builtins/logic_units/api_logic_unit.py
```diff
@@ -0,0 +1,69 @@
...
+from eidolon_ai_sdk.util.schema_to_model import schema_to_model
+
+
+class ApiLogicUnitSpec(BaseModel):
+    title: str = Field(description="Title of the API")
+    root_call_url: str = Field(description="Root URL of the API to call")
+    open_api_location: str = Field(description="Location of the OpenAPI schema")
+    operations_to_expose: List[Operation] = Field(description="Operations to expose")
+    extra_header_params: Optional[dict] = Field(description="Extra header parameters to add to every call. This can be a jinja template where the variables in the template are ENV variables (matching case)", default=None)
+    extra_query_params: Optional[dict] = Field(description="Extra query parameters to add to every call. This can be a jinja template where the variables in the template are ENV variables (matching case)", default=None)
```
LukeLalor, line 22: same nit down here

LukeLalor (approved):

PR #489, opened 2024-06-20
Azure file store

dbrewster:


sdk/pyproject.toml
```diff
@@ -97,3 +100,7 @@ develop = true
 [tool.poetry.group.dev.dependencies.eidolon-ai-usage-client]
 path = "../usage-service/usage-client"
 develop = true
+
+[tool.poetry.group.speedups.dependencies]
+aiohttp = "^3.9.5"
```
LukeLalor, line 105: what is the speedup group for?

sdk/pyproject.toml
```diff
@@ -69,6 +69,9 @@ posthog = "^3.5.0"
 ollama = "^0.2.0"
 pymupdf = "^1.24.5"
 jsonpath-ng = "^1.6.1"
+azure-storage-blob = "^12.20.0"
+aiohttp = "^3.9.5"
+aiodns = "^3.2.0"
```
LukeLalor, line 74: I don't see aiohttp or aiodns being used, do they need to be in the toml?

sdk/eidolon_ai_sdk/memory/chroma_vector_store.py
```diff
@@ -71,10 +74,12 @@ async def start(self):
         pass
 
     def connect(self):
+        print("******** connecting")
```
LukeLalor, line 77: missed these prints

sdk/pyproject.toml
LukeLalor: this group can disappear since it is in the main dependencies

LukeLalor (approved):

PR #542, opened 2024-07-17
React packaging

dbrewster:


jahabeebs:
@dbrewster I fixed the playwright test with the following changes
1) I noticed you renamed conversational_agent to conversational-agent in certain places. However when creating a new chat we were getting a 404 so I reverted this renaming back to conversational_agent and we were able to create new chats again 2) Our playwright test was failing because I was using a locator that was relying on the ID of the div of a chatbot's chat response being a certain number, but I made that test more specific (we now select by classname) and I also made sure that we match "Hello!" exactly as the chatbot's response

As a side note it seems now *when a user is logged in* the chatbot starts off with a blank message and the chatbot then responds with "Hello! Are you there?" right off the bat for some reason. This was contributing to the failing tests. Not sure if we want to fix this in this PR or a separate one, but the playwright test passes regardless of whether the user is signed in now

Screenshot 2024-07-18 at 12 59 36 AM

PR #612, opened 2024-08-12
First draft of long term memory with mem0

sidHathi:
Includes the new LongtermMemoryUnit class, integration into ConverstioalAPU (untested), and tests for the new memory unit.

**Notes**:
- The tests do not make use of the cassette system used by other tests - I need help figuring out how to make that work
- ConversationalAPU integration isn't tested yet - not sure if it's what we're going for
- Obviously this is my first PR for Eidolon and I'm sure I screwed some things up so would appreciate as much feedback as possible


sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def storeMessage(self, call_context: CallContext, message: LLMMessage):
+        user_id = AgentOS.current_user().id
+        agent_name = AgentOS.current_agent_name()
+        metadata = {
+            "process_id": call_context.process_id,
+            "agent_name": agent_name
+        }
+        # not sure if I should manipulate result somehow - what's the expected return value?
+        return self.mem0.add(message, user_id=user_id, metadata=metadata)
```
dbrewster, line 48: Should we be converting this to a string or does Luke's stuff handle it?

sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def storeMessages(self, call_context: CallContext, messages: List[LLMMessage]):
+        user_id = AgentOS.current_user().id
+        agent_name = AgentOS.current_agent_name()
+        metadata = {
+            "process_id": call_context.process_id,
+            "agent_name": agent_name
+        }
+        # not sure if I should manipulate result somehow - what's the expected return value?
+        return self.mem0.add(messages, user_id=user_id, metadata=metadata)
```
dbrewster, line 58: ditto

sdk/eidolon_ai_sdk/apu/conversational_apu.py
```diff
@@ -70,6 +73,9 @@ def __init__(self, spec: ConversationalAPUSpec = None):
         self.io_unit = self.spec.io_unit.instantiate(**kwargs)
         self.memory_unit = self.spec.memory_unit.instantiate(**kwargs)
         self.llm_unit = self.spec.llm_unit.instantiate(**kwargs)
+        # my best guess for how to initialize
```
dbrewster, line 76: This is correct

dbrewster (changes requested):
Se should go over a few things around scope but in general looks good

sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
sidHathi: i think mem0 itself handles it. i can convert to a string if that makes more sense, but it's worked this way so far

sdk/eidolon_ai_sdk/apu/conversational_apu.py
LukeLalor: we should remove this comment

sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
```diff
@@ -0,0 +1,146 @@
...
+from eidolon_ai_sdk.system.reference_model import Reference
+from eidolon_ai_sdk.system.reference_model import Specable
+
+
+class LongTermMemoryUnitScope(Enum):
+    SYSTEM = "system"
+    AGENT = "agent"
+    USER = "user"
+    USER_AGENT = "userAgent"
+    USER_PROCESS = "userProcess"
```
LukeLalor, line 22: You mentioned that you were going to change this? presumably by turning unit_scope into List[LongTermMemoryUnitScope] and allowing them to be used in conjunction? then restring the options to be user, agent, and process

sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
```diff
@@ -0,0 +1,146 @@
...
+
+        return list(filter(filter_func, memories))
+
+    def getMemory(self, memory_id: str):
+        return self.mem0.get(memory_id)
+
+    def getMemoryHistory(self, memory_id: str):
+        return self.mem0.history(memory_id)
+
+    def deleteMemory(self, memory_id: str):
```
LukeLalor, line 116: I don't see us using anything other than searchMemories and messages, do we need the other methods? Since this is the contract for long term memory, keeping it concise now will make expanding it easier.

sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, default_llm: LLMUnit, spec: LongTermMemoryUnitConfig = None,
+                 memory_converter: Optional[Callable[[List[ScoredPoint]], List[ScoredPoint]]] = None, **kwargs):
+        super().__init__(**kwargs)
+        self.spec = spec
+        self.unit_scope = spec.unit_scope
+        if spec is not None and spec.llm_unit is not None:
+            default_llm = spec.llm_unit
+        self.mem0 = EidolonMem0(default_llm, spec.collection_name, memory_converter=memory_converter)
+
+    def storeMessage(self, call_context: CallContext, message: LLMMessage):
```
LukeLalor, line 43: Let's make sure these methods are snake_case so the python snobs don't hate us

sdk/eidolon_ai_sdk/apu/conversational_apu.py
```diff
@@ -133,6 +139,12 @@ async def schedule_request(
             if self.record_memory:
                 await self.memory_unit.storeMessages(call_context, conversation_messages)
             conversation.extend(conversation_messages)
+
+            # EDIT: get relevant mem0 memories and extend the conversation history with them
+            if self.longterm_memory_unit:
+                longterm_memories = self.longterm_memory_unit.searchMemories(call_context, conversation_messages)
+                conversation.extend(longterm_memories)
```
LukeLalor, line 146: Does this work well just throwing the messages at the end?

I would have assumed we want to update the system message with the memories. I suppose this is something we can tweak with better evals and shouldn't block our initial impl merge.

LukeLalor (changes requested):
Looking good. Made some change requests in comments

sdk/eidolon_ai_sdk/apu/conversational_apu.py
sidHathi: Not sure if this should be an optional feature - should the user be able to specify whether they want to include the mem0 content as part of the system message? I agree that this customizability probably doesn't need to be part of the initial version

sdk/eidolon_ai_sdk/apu/longterm_memory_unit.py
sidHathi: i'll make that change

dbrewster (approved):

PR #630, opened 2024-08-21
Ll/simplify dev form

LukeLalor:


webui/apps/eidolon-ui2/app/eidolon-apps/dev-tool/components/start-program-dialog.tsx
```diff
@@ -5,7 +5,7 @@ import * as React from "react";
 import {usePathname, useRouter} from "next/navigation";
 import {ChooseAgentForm, getAppPathFromPath} from "@eidolon-ai/components/client";
 
-export function StartProgramDialog({open, onClose, machineUrl}: { machineUrl: string, open: boolean, onClose: (wasCanceled: boolean) => void }) {
+export function StartProgramDialog({open, onClose, machineUrl, defaultAgent}: { machineUrl: string, open: boolean, defaultAgent: string | undefined, onClose: (wasCanceled: boolean) => void }) {
```
steven-the-qa, line 8: Would it make sense to use `defaultAgent?: string` instead? Wondering if there's a use case for `undefined` here.

steven-the-qa (approved):
left a non-blocking question

webui/apps/eidolon-ui2/app/eidolon-apps/dev-tool/components/start-program-dialog.tsx
LukeLalor: No, you are right, this is better. I don't know typescript.

webui/README.md
```diff
@@ -53,12 +53,23 @@ The content are described in the template file.
 You will need to change the defaults if you would like to enable authentication or are running the eidos on
 a port other than 8080.
 
-### Running the Server
+### Running the Server for Development
+
+To build for dev mode, run:
+
+```bash
+turbo run build
```
jahabeebs, line 61: pnpm build--turbo is not a command that will be recognized unless they have it installed globally

webui/README.md
```diff
@@ -53,12 +53,23 @@ The content are described in the template file.
...
+```
+
+Or run a file watcher so that you do not need to rebuild when updating components.
+```bash
+turbo run watch
```
jahabeebs, line 66: watch is not in the package.json for webui--should the dev be running this from eidolon-ui-2?

webui/packages/eidolon-client/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "scripts": {
     "build": "tsup index.ts",
+    "watch": "tsup --watch",
```
jahabeebs, line 21: I see you added a watch command to the monorepos but ideally it would be added to webui because right now these have to be triggered individually--unless that's your intention

jahabeebs (changes requested):

webui/README.md
LukeLalor: build is defined in turbo.json, which I believe is where this command points to?

webui/README.md
LukeLalor: I added this command to turbo.json. I believe it picks up the dependencies transitively since it was working ootb since it was updating packages for me, but I definitely not sure if this is the right way to do it.

webui/packages/eidolon-client/package.json
LukeLalor: These guys are triggered by the root level `turbo run watch` command when I was running locally

jahabeebs (approved):

PR #638, opened 2024-08-23
add docs for how to swap llm

LukeLalor:


flynntsang:
Thanks @LukeLalor ! On a quick read, I think I can write additional copy to help first-time users. I'm going to walk through it and make sure I can successfully change my LLM before commenting further. 

I've been looking hard at the docs generally and how they line up (or don't) with the SDK. One thing that I'm coming around to is that agents can be of two kinds: APU-enabled agents or RAG agents. Am I getting that right?

LukeLalor:
Correct. There are other types as well, but that what we have docks for right now

flynntsang:
OK -- I was able to switch LLMs successfully, confirmed by changes to https://platform.openai.com/usage (and by asking the agent what version of chatgpt it is running).

A couple of things: 

1. s3_search.yaml is not an exclusive "RAG" agent. I discovered that it uses an APU (possibly for the question transformer?)  because I was still being charged for gpt-4-turbo after changing conversational_agent.yaml to use gpt-3.5-turbo.
2. Even though gpt-4o-mini is not listed as an LLMModel in the docs, I was able to use it. I wonder if there might be a better way to document the LLMModels for each LLM provider.

I'm going to think about ways to present this in the docs.

dbrewster (approved):

flynntsang (changes requested):
Can I take a crack at writing this for newbies? I've never been proficient with Github so I need some help with the mechanics.

LukeLalor:
@flynntsang definitely! I am going to merge this one so that it goes live, but please rip / revamp it!

PR #639, opened 2024-08-23
bad fix is the first step to good fix

LukeLalor:
Before
Screenshot 2024-08-23 at 12 31 01 PM

After
Screenshot 2024-08-23 at 12 31 11 PM


dbrewster (approved):
I tried a few better options but they are adding this in via a rule so it comes first unless we want to add layers into the css, which I do not

PR #674, opened 2024-08-29
pin spacy and remove build tools

LukeLalor:


Dockerfile
```diff
@@ -14,10 +14,9 @@ RUN poetry build
 FROM builder_base AS builder
 COPY sdk .
 RUN poetry build
-RUN poetry export --without dev --without-hashes --format=requirements.txt | grep -v "eidolon-ai-client" | grep -v "eidolon-ai-usage-client" | grep -v "-e file:///" > dist/requirements.txt
+RUN poetry export --without dev --format=requirements.txt | grep -v "eidolon-ai-client" | grep -v "eidolon-ai-usage-client" | grep -v "-e file:///" > dist/requirements.txt
```
LukeLalor, line 17: Although this is a hair slower, we need hashes to require the pinned version, otherwise we can't guarente we have tested against the versions on our image

dbrewster (approved):

PR #678, opened 2024-08-29
Mf/docs

flynntsang:


LukeLalor (changes requested):
We should not need to install the clients to use the different llm providers, so I do not think this is required documentation

flynntsang:
@dbrewster could you take a look please? This is a proposal to add a new How to, specifically how to authenticate with your LLM providers.

webui/apps/docs/astro.config.mjs
```diff
@@ -74,11 +74,12 @@ export default defineConfig({
...
-            {label: 'Agent-Agent Communication', link: '/docs/howto/communication'},
+            {label: 'Authenticate LLMs', link: '/docs/howto/authenticate_llm'},
             {label: 'Swap out LLMs', link: '/docs/howto/swap_llm'},
-            {label: 'Customizing Builtin Components', link: '/docs/howto/customize_builtins'},
-            {label: 'Custom Agent Templates', link: '/docs/howto/custom_agents'},
-            {label: 'Custom Components', link: '/docs/howto/using_references'},
+            {label: 'Configure Agent-Agent Communication', link: '/docs/howto/communication'},
+            {label: 'Configure Built-in Components', link: '/docs/howto/customize_builtins'},
+            {label: 'Use References', link: '/docs/howto/using_references'},
+            {label: 'Build Custom Agent Templates', link: '/docs/howto/custom_agents'},
```
dbrewster, line 82: I like the new name.  Should we consider foo - bar - bar instead

dbrewster (dismissed):

LukeLalor (approved):
ship it

PR #682, opened 2024-08-30
Changed headings to follow "How to..."

flynntsang:
OK -- I think I'm getting the basics figured out. I just made minor changes to the headings under the How To section.

### DIFF
--- a/webui/apps/docs/src/content/docs/docs/howto/communication.md
+++ b/webui/apps/docs/src/content/docs/docs/howto/communication.md
@@ -1,6 +1,6 @@
 ---
-title: Agent-Agent Communication
-description: References - Agent Communication
+title: Enable Agent Interactivity
+description: References - How to configure agents to communicate and interact with each other
 ---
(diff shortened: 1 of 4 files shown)


webui/apps/docs/src/content/docs/docs/howto/communication.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Agent-Agent Communication
-description: References - Agent Communication
+title: Enable Agent Interactivity
```
LukeLalor, line 2: nit: it is always enabled, but this shows how to utilize it. Utilizing Agent Interactivity?

LukeLalor (changes requested):
Text displayed in the sidebar is at webui/apps/docs/astro.config.mjs

this just changes the title when you open the article.

Title and sidebar link can be different, and I think they probably should be.

For example, `how to... use references` makes a lot of sense in the sidebar, but if you were linked to the page from somewhere else, the title `using references` makes much more sense in isolation than `Use references`. The alternative would also be to prefix each title, so it would be `How to Use References` which is another option

PR #687, opened 2024-09-03
add latest models

LukeLalor:


sdk/eidolon_ai_sdk/builtins/resources/claude_sonnet.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   image_unit: OpenAIImageUnit
   llm_unit:
     implementation: AnthropicLLMUnit
-    model: "claude-3-sonnet-20240229"
+    model: "claude-3-5-sonnet-20240620"
```
dbrewster, line 14: Do we want to change the existing one or add a new? This will fail if people are still using Sonnet 3

dbrewster: NVM.  I'm an idgot

dbrewster (approved):

PR #689, opened 2024-09-04
pull out playwright and optomize image. size ~2gb -> 777mb

LukeLalor:


dbrewster:
Can we add this to the web researcher project when this goes in?

dbrewster (approved):

PR #693, opened 2024-09-04
Added same-id tracing to UI, bakend and docs

dbrewster:


webui/apps/docs/src/components/common/posthog.astro
```diff
@@ -2,11 +2,13 @@
...
   posthog.init(
     'phc_9lcmDyxVkji98ggIqy2XvyVcItnrgdrMQhZBFp6Du5d',
     {
       api_host:"https://us.posthog.com"
     }
   )
+  posthog.identify(md5(window.location.hostname));
```
LukeLalor, line 13: @dbrewster isn't this just the url? won't this be the same for everybody?

dbrewster: Damn copilot!!!!

LukeLalor (approved):

PR #695, opened 2024-09-04
web docs redirects

### LINKED ISSUE #691: Broken Docs Pages
https://www.eidolonai.com/docs/howto/customize_builtins
https://www.eidolonai.com/docs/components/simple_agent#defining-actions

flynntsang:
should fix #691

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Now that you have a running agent machine with a simple agent. Let's start custo
 
 - [ ] Add new capabilities via logic units (tools)
 - [ ] Enable [agent-to-agent communication](https://www.eidolonai.com/docs/howto/communication)
-- [ ] [Swap out components](https://www.eidolonai.com/docs/howto/customize_builtins) (like the underlying llm)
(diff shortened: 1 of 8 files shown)


LukeLalor (changes requested):
should we find those old links and fix them instead?

flynntsang:
> should we find those old links and fix them instead?

We should, but it's very common to have long redirects. There might be links in emails and who-knows-what that we don't have control over.

LukeLalor:
> > should we find those old links and fix them instead?
> 
> We should, but it's very common to have long redirects. There might be links in emails and who-knows-what that we don't have control over.

adding the redirects is fine, but we should also fix the links in 
what_is_agent.mdx
README.md

flynntsang:
@LukeLalor Please give the files another look-over.

@abhi-vachani I was in the quickstart.md file and made the changes from hello_world to hello-world where appropriate.

LukeLalor (approved):

PR #721, opened 2024-09-09
Mmf/docs quickstart

flynntsang:
quickstart updates; open external links in new tab; fixed broken links


webui/apps/docs/src/content/docs/docs/quickstart.md
```diff
@@ -12,106 +12,143 @@ description: Walk through prequisites for Mac and Linux
...
+## Change the System Prompt
 
-Did your agent respond to you? 🍾 Congratulations! You have successfully created your first agent machine.
+To get a taste for how quickly you can deploy powerful agentica applications, make a simple change.
+
+Using a text editor, change the system prompt:
+
+```yaml title=resources/hello-world.yaml
+  system_prompt: |
+    Translate {{message}} to Spanish.
```
LukeLalor, line 146: Can you update this to something akin to...

You are a translator who translates user messages into Spanish. When a user sends you a message, respond with the translation of their message in Spanish.

webui/apps/docs/src/content/docs/docs/quickstart.md
```diff
@@ -12,106 +12,143 @@ description: Walk through prequisites for Mac and Linux
...
 
-This file describes how to instantiate your agent from its **AgentTemplate** 🏭 and describes any customization you might
-want (like a custom LLM. tools, etc).
-
-```yaml
-apiVersion: eidolon/v1
-kind: Agent
-metadata:
-  name: hello-world
+1. Open a web browser to http://localhost:3000 (or replace localhost with your server URL).
```
LukeLalor, line 81: we can send them directly to the dev tool link to skip a step if you think that would be a better UX. I don't have an opinion here, just pointing out the possibility.

LukeLalor (changes requested):
looks good! One suggestion on the spanish translation since I don't think the agent will have an understanding of {{ message }} and I worry about how the jinja template would get formatted.

webui/apps/docs/src/content/docs/docs/quickstart.md
flynntsang: What about: Translate {{body}} to Spanish? I misremembered the field from http://localhost:4321/docs/components/agents/simpleagent#actions. I think we should be able to reference {{body}} no? @LukeLalor

webui/apps/docs/src/content/docs/docs/quickstart.md
flynntsang: Valid point. From a learning perspective, I think it's good to see this step.

webui/apps/docs/src/content/docs/docs/quickstart.md
flynntsang: @LukeLalor -- Can you try a conversation at http://ec2-54-242-183-82.compute-1.amazonaws.com:3000? 

Here's my updated prompt, which so far is working great. I'm a sucker for brevity.

```yaml
  system_prompt: |
    Repeat {{body}}, than translate {{body}} into Spanish.
```
Is there a way to check how the jinja template is formatting this, to make sure it's all ok?

LukeLalor: you can use {{ body }} in the user prompt, but not the system prompt. There is no templating at the system prompt

flynntsang:
Changed the prompt (but did keep it short). Added images!

LukeLalor (approved):
ship it!

Issue #726, opened 2024-09-09
Agent controller doesn't respect namespace


dbrewster:
The agent controller is always creating resources in the default namespace and not obeying the namespace passed to it.

PR #728, opened 2024-09-09
Fixed agent operator to publish to the correct namespace

dbrewster:
Issue #726


LukeLalor (commented):
similar to masons comment, could you add a test for the default namespace case?

LukeLalor (approved):

PR #763, opened 2024-09-12
Fix output of simple agent events

dbrewster:


dbrewster:
Changes look good, but you have to approve it Luke.  I created it.

PR #779, opened 2024-09-13
Ll/strawberry blog

LukeLalor:


mfernest:
I put my comments in the branch. To summarize, I find it well-written and energetic in tone. Maybe brag a bit more on how 'easy' it was to adapt. To an experienced developer I'm thinking that story is even more compelling than the convenience you've implemented.

Maybe you could also say something about backwards compatibility? As in how/why this change doesn't break an implementation for other models.

mfernest (approved):

PR #783, opened 2024-09-16
Ll/vectara

LukeLalor:
Add vectara support in Eidolon.

1. Adds vectara logic unit for simple agentic rag
2. Adds vectara agent utilizing built in chat interface. 

Will add docs in follow-up pr if we like the contract


sdk/eidolon_ai_sdk/agent/vectara_agent.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+os.environ.setdefault("VECTARA_API_KEY", "test")
+
+
+class VectaraAgentSpec(BaseModel):
+    corpus_key: str
+    description: str = "Search documents related to {{ corpus_key }}"
+    vectara_url: str = "https://api.vectara.io/"
+    body_overrides: dict = Field({}, description="Arguments to use when creating / continuing a chat. See https://docs.vectara.com/docs/rest-api/create-chat for more information.")
```
LukeLalor, line 26: Agent Configuration lives here. Override allow the creation / turn interface to be arbitrarily customized since we don't know how people will want to use This. contract is the same for both endpoints,

sdk/eidolon_ai_sdk/builtins/logic_units/vectara.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class VectaraSearchSpec(BaseModel):
+    """
+    A logic unit for searching in Vectara. Requires the VECTARA_API_KEY environment variable to be set for authentication.
+    """
+
+    corpus_key: str = Field(description="The corpus key to search in.")
+    description: str = Field("Search documents related to {corpus_key}.", description="Description of the tool presented to LLM. Will be formatted with corpus_key.")
+    vectara_url: str = "https://api.vectara.io/"
```
LukeLalor, line 18: Vectara tool for agentic rag. Allows agents to 1. query arbitrary search terms and 2. read full documents as desired (doc id is retrieved from searches)

Issue #709, opened 2024-09-05
Unfunded OpenAI account does not produce easy to understand errors


abhi-vachani:
Problem: 
- When trying to converse with an agent in quickstart, if one does not have a funded OpenAI account, unclear error occurs: "Agent transitioned to state: unhandled_error"
- Error Trace: agent-server-1  | openai.NotFoundError: Error code: 404 - {'error': {'message': 'The model `gpt-4-turbo` does not exist or you do not have access to it.', 'type': 'invalid_request_error', 'param': None, 'code': 'model_not_found'}}
agent-server-1  | 
agent-server-1  | The above exception was the direct cause of the following exception:
agent-server-1  | 
agent-server-1  | Traceback (most recent call last):
agent-server-1  |   File "/usr/local/lib/python3.11/site-packages/eidolon_ai_sdk/system/agent_controller.py", line 346, in stream_agent_iterator
agent-server-1  |     async for event in stream:
agent-server-1  |   File "/usr/local/lib/python3.11/site-packages/eidolon_ai_sdk/agent/simple_agent.py", line 238, in fn
agent-server-1  |     async for e in action_fn(self, action, process_id, **kwargs):
agent-server-1  |   File "/usr/local/lib/python3.11/site-packages/eidolon_ai_sdk/agent/simple_agent.py", line 282, in _act
agent-server-1  |     async for event in response:
agent-server-1  |   File "/usr/local/lib/python3.11/site-packages/eidolon_ai_sdk/apu/conversational_apu.py", line 144, in schedule_request
agent-server-1  |     raise APUException(f"{e.__class__.__name__} while processing request") from e
agent-server-1  | eidolon_ai_sdk.apu.apu.APUException: APU Error: NotFoundError while processing request

Solution:
Display clear error message explaining problem is due to unfunded OpenAI account.

flynntsang:
FYI, I got this error with Anthropic when my free credits ran out:

`Error
Anthropic Status Error: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'Your credit balance is too low to access the Claude API. Please go to Plans & Billing to upgrade or purchase credits.'}}
`

coryfoo:
[...]

PR #784, opened 2024-09-17
OpenAI request `NotFoundError` response adds additional context to indicate potential lack of funds

coryfoo:
When an OpenAI token is used that either does not have billing enabled or does not have enough credits to call the default model, their API will raise an exceptionof type `NotFoundError`. 

When this happens, we will catch the errors of this type and modify the message to include additional context that indicates that the issue may be related to insufficient credits. We will then re-raise an `APUException`.

Fixes issue #709


coryfoo:
When looking into this issue, I also considered a potentially more robust implementation that checked for the available OpenAI models at server start time and halted (or at least logged pretty aggressively). I ultimately decided against that "belt and suspenders" approach as this is a really a transitory issue that could easily be remedied without a server restart. That, and as I am new to this project, I was not sure how welcome a change would be that veers outside the scope declared within the issue.

LukeLalor:
thanks @coryfoo! This will already be a much better error.

Could you add an api level test to verify the response? `sdk/tests/test_sql_agent.py` is a pretty tight example for how to define a test module that starts the server with a custom agent definition. Obviously you would need to use the SimpleAgent instead of the SqlAgent, but you get the idea.

Another potential addition is is to make a request request to https://platform.openai.com/docs/api-reference/models/list in the catch block so that we can disambiguate bad model vs unfunded account. We would need to guard against errors here, but it could allow us to be very helpful in most cases.

coryfoo:
> Could you add an api level test to verify the response? `sdk/tests/test_sql_agent.py` is a pretty tight example for how to define a test module that starts the server with a custom agent definition. Obviously you would need to use the SimpleAgent instead of the SqlAgent, but you get the idea.

I'd be happy to do that! Is there an existing test file that makes sense to dump the test into, or should I create a new one?

LukeLalor:
> > Could you add an api level test to verify the response? `sdk/tests/test_sql_agent.py` is a pretty tight example for how to define a test module that starts the server with a custom agent definition. Obviously you would need to use the SimpleAgent instead of the SqlAgent, but you get the idea.
> 
> I'd be happy to do that! Is there an existing test file that makes sense to dump the test into, or should I create a new one?



> > Could you add an api level test to verify the response? `sdk/tests/test_sql_agent.py` is a pretty tight example for how to define a test module that starts the server with a custom agent definition. Obviously you would need to use the SimpleAgent instead of the SqlAgent, but you get the idea.
> 
> I'd be happy to do that! Is there an existing test file that makes sense to dump the test into, or should I create a new one?



> > Could you add an api level test to verify the response? `sdk/tests/test_sql_agent.py` is a pretty tight example for how to define a test module that starts the server with a custom agent definition. Obviously you would need to use the SimpleAgent instead of the SqlAgent, but you get the idea.
> 
> I'd be happy to do that! Is there an existing test file that makes sense to dump the test into, or should I create a new one?



> > Could you add an api level test to verify the response? `sdk/tests/test_sql_agent.py` is a pretty tight example for how to define a test module that starts the server with a custom agent definition. Obviously you would need to use the SimpleAgent instead of the SqlAgent, but you get the idea.
> 
> I'd be happy to do that! Is there an existing test file that makes sense to dump the test into, or should I create a new one?

You can create a new test file. Most of our other tests are parameterized by llm provider, and we just want it for openai.

coryfoo:
@LukeLalor I made a new test file, per your request. I didn't exactly copy the `test_sql_agent.py` pattern, however, as my feeling was that the goal was to assert that the error message was appropriate when the OpenAI client threw a `NotFoundError`. Therefore, I mocked out a new subclass of `OpenAIConnectionHandler` so that we could make it raise the error. I felt that going the route of starting a server and configuring an agent was a less direct path of accomplishing this goal.

LukeLalor:
sounds good. Thanks for that @coryfoo 

failing lock file test is due to external contributors bug. I filed it at https://github.com/eidolon-ai/eidolon/issues/794 and will merge bypass the rule.

PR #785, opened 2024-09-17
no-tests-needed to add contributors

LukeLalor:


abhi-vachani (commented):
Thanks for asking me to review this so I could get more familiar with the codebase -- is the point of these changes to allow public contributors to not have to pass the automatic tests in place when creating a pr?

LukeLalor:
> Thanks for asking me to review this so I could get more familiar with the codebase -- is the point of these changes to allow public contributors to not have to pass the automatic tests in place when creating a pr?



> Thanks for asking me to review this so I could get more familiar with the codebase -- is the point of these changes to allow public contributors to not have to pass the automatic tests in place when creating a pr?

correct, it does not impact the test runs, so there is no point running the tests when those files change.

abhi-vachani (approved):
Ok, awesome thanks. I'll approve this bc it seems like a simple change, but just curious then what are the github actions checking for / doing?

PR #788, opened 2024-09-17
create azure llm provider component

LukeLalor:
Screenshot 2024-09-18 at 2 26 47 PM


examples/eidolon_examples/azure_quickstart/resources/hello_world_agent.yaml
```diff
@@ -2,8 +2,15 @@
 apiVersion: server.eidolonai.com/v1alpha1
 kind: Agent
 metadata:
-  name: hello_world
+  name: hello-world
```
dbrewster, line 5: Did you mean to do this? Are you automatically doing the conversion from - to _ now?

dbrewster (changes requested):

examples/eidolon_examples/azure_quickstart/resources/hello_world_agent.yaml
LukeLalor: I did mean to do this. We should be moving to - instead of _ and eventually enforcing it in the object parsing

PR #790, opened 2024-09-19
Massive web cleanup.

dbrewster:
Massive cleanup to the webUI. 

First some architectural changes:
  * Removed material design (mui) and mii-icons
  * Replaced with tailwind and lucide
  * Added SIMPLE hooks for getApp, getProcess, and getProcesses. The getProcess has an update hook on it so the UI can update when it needs to (for things like background title changes). This greatly simplified the components.
  * Made dev tool and chatbot tool more inline with each other. Only divergence now is the main page (for each app) and the input panel.

Python changes:
  * Added / cleaned up support for title and subtitle on agents and tools. You can now specify the title / subtitle for each and they appear in the UI correctly.
  * Cleaned up when isAgent is set on events.
  * Added a "hidden" flag to object output and string output events.  Idea is these are events internal to the app and shouldn't be shown to a user.
  * Made a change to simple agent where it now generates the title automatically for you, in the background, and sets the process title correctly.

General LAF changes:
  * Header is now more inline with the eidolonai website and removed login button. 
New Header

  * Home page is cleaned up.
Home Page

  * Sidebar can now float
Floating sidebar

  * or be pinned
Pinned sidebar

  * Output pane has been cleaned up with lots of changes to how markdown is rendered. + a new font.
Discussion Page

  * Sub-agent calls now can be drilled down on to enter the sub-agent conversation (read-only)
Sub-agent actions

  * Conversation pages now have a new header.
  * New Header
  * And shows breadcrumbs when you drill down
Breadcrumbs

  * Scroll region shows fade and scroll locking (i.e. when you scroll to the bottom of the page) was reimplemented and works now.
New Scroll Fade

Chatbot LAF changes:
  * The landing page is new. It now lists instructions and descriptions of the bot with the ability to input text on that page
Landing Page
.
  * File upload is redesigned and doesn't upload the files until the first command is sent.
File Upload

  * The input panel is completely redesigned to be smaller and with new buttons / info text
New Panel Buttons

Devtool LAF changes:
  * The landing page is completely redesigned showing the agents and operations, with their titles and descriptions, available on the page.
Landing Page

  * The input panel is also redesigned and now works much, much better.
  
Devtool Input #1
Devtool Input #2


webui/apps/eidolon-ui2/app/eidolon-apps/chatbot/app_manifest.json
```diff
@@ -11,7 +11,7 @@
     "inputLabel": "How can I help you?",
     "titleOperationName": "generate_title",
     "allowSpeech": true,
-    "speechAgent": "speech-agent",
+    "speechAgent": "speech_agent",
```
LukeLalor, line 14: rename the agent with a -, we can't use _ in k8's names so need to start migrating

LukeLalor (changes requested):
Some problems this creates with the quickstart:

1. It is now too many clicks to use the hello world agent (3), we should expand the first agent in the list.
2. Converse (the action name) is duplicated. Why is this? is it grabbing the action description? if so this can be updated in the quickstart, not this pr
3. [object, Object] text indicates that it does not handle plain/text body types well
4. defaulting to the same input as the previous request is probably not desired behavior


Screenshot 2024-09-25 at 8 16 25 AM
Screenshot 2024-09-25 at 8 16 31 AM
Screenshot 2024-09-25 at 8 16 42 AM
Screenshot 2024-09-25 at 8 15 31 AM

client/python/eidolon_ai_client/events.py
```diff
@@ -122,6 +123,7 @@ class StringOutputEvent(OutputEvent):
 class ObjectOutputEvent(OutputEvent, Generic[T]):
     event_type: Literal["object"] = "object"
     content: T
+    hidden: Optional[bool] = False
```
LukeLalor, line 126: hidden becomes very ui specific. Perhaps it would be better to add a metadata object to any event, and then we can throw this in there?

If not, this shouldn't be Optional[bool], just make it a bool

sdk/eidolon_ai_sdk/system/agent_controller.py
```diff
@@ -271,12 +273,11 @@ async def agent_event_stream(self, handler, process, last_state, **kwargs) -> As
         is_async_gen = inspect.isasyncgenfunction(handler.fn)
         stream = handler.fn(self.agent, **kwargs) if is_async_gen else self.stream_agent_fn(handler, **kwargs)
 
-        def get_start_event():
+        async def get_start_event():
```
LukeLalor, line 276: do we need to make this extra mongo call? haven't we just retrieved the process doc?

client/python/eidolon_ai_client/events.py
dbrewster: How about "internal"? There are, and will likely be more of, events we don't want any UI to render, like this one.

dbrewster:
I didn't everything but hidden -> internal.  LMK

webui/apps/eidolon-ui2/app/eidolon-apps/chatbot/app_manifest.json
flynntsang: I think the original has the hyphen

LukeLalor:
looks like links do not show up properly in the markdown. They inherit color / style so they are not blue/purple/underlined
Screenshot 2024-09-25 at 6 00 16 PM

LukeLalor:
We should probably also update the 
quickstart docs [https://www.eidolonai.com/docs/quickstart](https://www.eidolonai.com/docs/quickstart) to remove the old screenshots. We can update it later with a gif or something to make it clear/concise, but that can be a separate pr imo

client/python/eidolon_ai_client/events.py
LukeLalor: if that is the case, shouldn't we just put it in a separate context? maybe even a context with the name eidolon-hidden which the ui treats differently? seems like something we can represent without adding a new concept.

webui/apps/eidolon-ui2/app/eidolon-apps/chatbot/app_manifest.json
LukeLalor: correct. It needs the hyphen, not the _

dbrewster: The file checked in has an underscore in it now

LukeLalor (approved):

PR #792, opened 2024-09-23
Update authenticate_llm.md

flynntsang:
Reminder to restart after editing .env

### DIFF
--- a/webui/apps/docs/src/content/docs/docs/howto/authenticate_llm.md
+++ b/webui/apps/docs/src/content/docs/docs/howto/authenticate_llm.md
@@ -49,4 +49,4 @@ MISTRAL_API_KEY=yourmistralapikey
 
 Restart Eidolon. Agents can now authenticate with your preferred LLMs.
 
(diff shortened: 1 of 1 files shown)


webui/apps/docs/src/content/docs/docs/howto/authenticate_llm.md
```diff
@@ -49,4 +49,4 @@ MISTRAL_API_KEY=yourmistralapikey
 
 Restart Eidolon. Agents can now authenticate with your preferred LLMs.
 
->Note: if the `.env` file does not exist in your application root directory, create it as a plain text file. 
+>Note: if the `.env` file does not exist in your application root directory, create it as a plain text file. Always restart after editing the `.env` file.
```
LukeLalor, line 52: We should add this to our file watcher so that the server automatically restarts when in watch mode.

I'll file an issue for that, and then we can revert this change when that goes in.

LukeLalor (approved):

PR #830, opened 2024-10-01
title_generation_mode from auto to none

abhi-vachani:


LukeLalor:
@abhi-vachani the change looks good. You will need to update the failing tests that depend on the title generation mode being set by default, but other than that looks good.

also looks like docs up to date cannot auto commit for external contributions. You can run the doc script locally to make that test pass.

LukeLalor (approved):

dbrewster (approved):

PR #831, opened 2024-10-01
Basic Gemini Support

coryfoo:
Adds basic support for the Google Gemini LLM (`gemini-1.5-flash`). Does not include Gemini tool support for multi-modal interactions using Gemini-based models (ie, no speech, pictures, etc).

I've taken the code from the [existing branch](https://github.com/eidolon-ai/eidolon/tree/eab-gemini) and updated it so that the protobuf serialization works, specifically when declaring tool support. You are able to interact with a chat-only version of this implementation which works fine. 

The tool calling portions are untested (could use some help with setting this up locally!).


coryfoo:
The documentation for image and audio uploads seems to suggest that it does not support one-shot model prompts that include the image data embedded within the content as other models do, but rather you must upload a file and refer to that upload in the prompt. Because of that, I wasn't sure of the best approach to take to add support for that paradigm and would be happy to chat about potential implementation strategies @dbrewster @LukeLalor ?

PR #835, opened 2024-10-08
Changed the client id environment variable

dbrewster:


sdk/eidolon_ai_sdk/security/azure_authorizer.py
```diff
@@ -11,7 +11,7 @@
 
 class AzureJWTProcessorSpec(BaseModel):
     client_id: str = Field(
-        os.environ.get("AZURE_CLIENT_ID") or os.environ.get("AZURE_AD_CLIENT_ID"),
```
LukeLalor, line 14: did you mean to do this? I don't think this is what azure uses by default

https://learn.microsoft.com/en-us/dotnet/api/azure.identity.environmentcredential?view=azure-dotnet

LukeLalor (changes requested):

Issue #876, opened 2024-10-28
[ problem ] Vectara validation error 


flynntsang:
**Describe the problem**
When running the Vectara logic unit:
`APU Error: ValidationError while processing request`

**Agent YAML file**
```yaml
apiVersion: server.eidolonai.com/v1alpha1
kind: Agent
metadata:
  name: riverparkresearch
spec: 
  implementation: SimpleAgent
  apu: 
    implementation: ConversationalAPU
    logic_units:
      - implementation: VectaraSearch
        corpus_key: riverpark-chat
```
Log file attached
[logs_vectara_tool_validation_error.txt](https://github.com/user-attachments/files/17547517/logs_vectara_tool_validation_error.txt)

PR #878, opened 2024-10-29
fix vectara bug + add tests

LukeLalor:


LukeLalor:
[#876](https://github.com/eidolon-ai/eidolon/issues/876)

PR #882, opened 2024-10-31
better docs for local file memory

LukeLalor:


flynntsang (commented):
Would https://docs.docker.com/engine/storage/volumes/#use-a-volume-with-docker-compose be a better reference for docker, considering that we ship with a docker-compose.yml file?

flynntsang (commented):
Also, I think we need these edits for /components/documentloader/filesystemloader in addition to localfilememory.

flynntsang:
> > Also, I think we need these edits for /components/documentloader/filesystemloader in addition to localfilememory.
> 
> We can remove the default filesystemloader such that the document loader needs to be specified, but I don't think anything besides a note that the document loader defines where the files will be ingested from and needs to be properly configured is appropriate. Is that what you are asking for?

I think we need to be more helpful than just saying "Root Dir".  If we want to help less experienced devs be successful with complex AI, we need to do a little more empathetic and point them in the right direction. I don't want our evaluators to be frustrated or feel inadequate because they can't figure out why they can't get to their files.

image

flynntsang:
> > Would https://docs.docker.com/engine/storage/volumes/#use-a-volume-with-docker-compose be a better reference for docker, considering that we ship with a docker-compose.yml file?
> 
> Changed it over to this. The main goal is to remind the dev that in a containerized environment their file system will be different.

Do we need both of these? I think for my example we only used the second link?

- [Build the files into the runtime image](https://docs.docker.com/reference/dockerfile/#copy)
- [Docker volume](https://docs.docker.com/engine/storage/volumes/#use-a-volume-with-docker-compose)

LukeLalor:
> > > Would https://docs.docker.com/engine/storage/volumes/#use-a-volume-with-docker-compose be a better reference for docker, considering that we ship with a docker-compose.yml file?
> > 
> > 
> > Changed it over to this. The main goal is to remind the dev that in a containerized environment their file system will be different.
> 
> Do we need both of these? I think for my example we only used the second link?
> 
> * [Build the files into the runtime image](https://docs.docker.com/reference/dockerfile/#copy)
> * [Docker volume](https://docs.docker.com/engine/storage/volumes/#use-a-volume-with-docker-compose)

Correct, but there are a ton of ways devs could choose to make the files accessible. For real deployments, it likely mean building it into the image is one of the more likely choices, so that one should stay. Our goal needs to be just as much "how to deploy" as "how to run locally"

PR #900, opened 2024-11-08
fix titles

LukeLalor:


flynntsang (commented):
Should we append APU to the (APU) titles here as well?

flynntsang (commented):
lgtm

PR #905, opened 2024-11-12
use same test infrastructure for example tests

LukeLalor:


LukeLalor:
Only test failing is the flakey test that this is to help debug, so I am going to merge this

PR #908, opened 2024-11-12
it is helpful to actually respond with information about the error th…

LukeLalor:
…at just happened. This gets saved into the history, which we may or may not want


LukeLalor:
Screenshot 2024-11-11 at 8 53 18 PM

LukeLalor:
not sure if this will break tests or not, but we should consider responding with details when an error happens. this allows us to actually give errors to the user when we are streaming rather than just closing the connection (which is not super helpful)

LukeLalor:
Closing this, the error is already recorded with state change

PR #931, opened 2024-11-19
[no-merge] Ll/853/efficient git

LukeLalor:


LukeLalor:
evidently there is a test failing with mongo memory? Todo investigate

